PR #9, opened 2019-12-03
[Docs] Refer to individual plugins on doc landing page

robbavey:


robbavey:
@karenzone - Would you mind looking at the docs for the JDBC integration plugin? This includes this PR to create a landing page, and the general structure of the docs pages underneath, and make sure that it conforms

docs/index.asciidoc
```diff
@@ -21,6 +21,10 @@ include::{include_path}/plugin_header.asciidoc[]
 
 ==== Description
 
-The JDBC Integration Plugin provides integrated plugins for working with Databases.
+The JDBC Integration Plugin provides the following plugins for working with databases that provide a JDBC interface:
```
karenzone: I'd like to standardize on the wording used for integrations.  Something like:

"The JDBC Integration Plugin provides integrated plugins for working with https://www.oracle.com/technetwork/java/javase/jdbc/index.html[JDBC]."

karenzone: If you decide to include a link, please verify that the sample link I provided is the best resource before using it.

robbavey: @karenzone Is it OK to not include a link? I'm not sure it adds a huge amount of value - it mentions a few different Java database technologies, with a dead link to a JDBC overview...  

Does something like this work?

"The JDBC Integration Plugin provides integrated plugins for working with Databases that provide JDBC drivers?"

karenzone: Absolutely the right call.  If we can't identify a definitive link, then let's skip it. What you are suggesting is fine, which one minor tweak. "Databases" should be lowercase.

robbavey:
@karenzone I pushed the change you requested. 

Also, is the rest of the doc structure in this plugin ok?

robbavey:
Thank you @karenzone!

PR #17, opened 2020-02-03
Fix tracking_column regression with Postgresql Numeric types

robbavey:
Prior to logsatash-input-jdbc 4.3.14, it was possible to use
the Postgresql Numeric type as a tracking column across Logstash
runs. Since 4.3.14, the deserialization of the sql_last_run
parameter has been broken for BigDecimal types, which is how
Numeric types are represented as Ruby BigDecimal objects by the
Sequel library we use in this plugin.

This commit restores this functionality.

Fixes: https://github.com/logstash-plugins/logstash-input-jdbc/issues/371

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/value_tracking.rb
+++ b/lib/logstash/plugin_mixins/jdbc/value_tracking.rb
@@ -62,7 +62,7 @@ module LogStash module PluginMixins module Jdbc
 
   class NumericValueTracker < ValueTracking
     def set_initial
-      common_set_initial(:gcd, 0)
+      common_set_initial(:abs, 0)
     end
 
     def set_value(value)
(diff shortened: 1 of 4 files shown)


lib/logstash/plugin_mixins/jdbc/value_tracking.rb
```diff
@@ -62,7 +62,7 @@ def common_set_initial(method_symbol, default)
 
   class NumericValueTracker < ValueTracking
     def set_initial
-      common_set_initial(:gcd, 0)
+      common_set_initial(:abs, 0)
```
andsel: I suspect the error before was due to the fact that BigDecimal hasn't gcd method.

andsel: And the `gcd(x, 0)` was a little bit trickie way to get the absolute value of x

andsel:
LGTM

robbavey:
Thanks @anndsel!

PR #20, opened 2020-03-05
Fix Integration Tests

robbavey:
Integration tests were failing due to a behavioural change in a
patch release of the postgres docker image, which now requires
a password to be set, or an explicit ENV setting to not require this.

This commit also fixes tests that were broken due to the changes to
devutils 2.0


robbavey:
Thanks @kares!

PR #24, opened 2020-03-24
customize testing files to account for integration test

jsvd:
because this plugin needs another container for an integration
test we're adding here the necessary tweaks for this setup:

'run.sh' will run both normal and integration tests
'setup.sh' will fetch the postgres driver from within the logstash container
'docker-compose.override.yml' adds the postgres container

This is blocked on https://github.com/logstash-plugins/.ci/pull/9 getting merged.

### DIFF
--- a/.ci/docker-compose.override.yml
+++ b/.ci/docker-compose.override.yml
@@ -1,0 +1,20 @@
+version: '3'
+
+services:
+
+  logstash:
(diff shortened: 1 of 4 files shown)


andsel:
LGTM after the other is  logstash-plugins/.ci#9 is merged and travis is green

PR #29, opened 2020-04-25
Fix jdbc_fetch_size with postgresql

bmax:
Port from old library to new: https://github.com/logstash-plugins/logstash-input-jdbc/pull/368

According to documentation (https://jdbc.postgresql.org/documentation/head/query.html) we need autocommit disabled. Work around is to create transaction and rollback always.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/statement_handler.rb
+++ b/lib/logstash/plugin_mixins/jdbc/statement_handler.rb
@@ -31,15 +31,19 @@ module LogStash module PluginMixins module Jdbc
     # @yieldparam row [Hash{Symbol=>Object}]
     def perform_query(db, sql_last_value, jdbc_paging_enabled, jdbc_page_size)
       query = build_query(db, sql_last_value)
-      if jdbc_paging_enabled
-        query.each_page(jdbc_page_size) do |paged_dataset|
-          paged_dataset.each do |row|
+      # Execute query in transaction cause PG driver require autocommit off for set fetch count
+      # See: https://jdbc.postgresql.org/documentation/head/query.html
+      db.transaction(rollback: :always) do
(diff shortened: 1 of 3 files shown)


spec/inputs/jdbc_spec.rb
```diff
@@ -1076,7 +1075,7 @@
     end
 
     it "should report the statements to logging" do
-      expect(plugin.logger).to receive(:debug).once
+      expect(plugin.logger).to receive(:debug).thrice
```
robbavey, line 1078: Do we know why this needed to change?

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -31,16 +31,20 @@ class NormalStatementHandler < StatementHandler
     # @yieldparam row [Hash{Symbol=>Object}]
     def perform_query(db, sql_last_value, jdbc_paging_enabled, jdbc_page_size)
       query = build_query(db, sql_last_value)
-      if jdbc_paging_enabled
-        query.each_page(jdbc_page_size) do |paged_dataset|
-          paged_dataset.each do |row|
+      # Execute query in transaction cause PG driver require autocommit off for set fetch count
+      # See: https://jdbc.postgresql.org/documentation/head/query.html
+      db.transaction(rollback: :always) do
```
robbavey, line 36: I wonder if it might be possible to potentially use [paged_each](https://www.rubydoc.info/github/jeremyevans/sequel/Sequel%2FDataset:paged_each) here? 

If not, do we need to use transactions when paging is not enabled?

robbavey (commented):
Thanks for the contribution @bmax - I have a couple of questions

PR #32, opened 2020-05-26
[Tests] Increase frequency of reload to fix flaky tests

robbavey:
JDBC static scheduled reload tests were flaky, due to scheduled
reload not happening every test, due to a mismatch between sleep
time and schedule.


robbavey:
@andsel The original was set to run every 10 minutes, rather than every 10 seconds - see https://www.elastic.co/guide/en/logstash/master/plugins-filters-jdbc_static.html#plugins-filters-jdbc_static-loader_schedule

spec/filters/integration/jdbc_static_spec.rb
```diff
@@ -136,14 +136,14 @@ module LogStash module Filters
...
-          settings["loader_schedule"] = "*/10 * * * * * UTC"
+          settings["loader_schedule"] = "*/3 * * * * * UTC"
           Timecop.travel(Time.now.utc - 3600)
           Timecop.scale(60)
           static_filter = JdbcStatic.new(settings)
           runner = Thread.new(static_filter) do |filter|
             filter.register
           end
-          sleep 3
+          sleep 4
```
yaauie, line 146:
```suggestion
          runner.join(4)
          sleep 4
```

andsel:
I could agree if the `*` after the first was 4 instead of 5, looking at Fugit (https://github.com/floraison/fugit#fugitnat) when 6 slots are provided the first is for seconds, or I'm looking something in the bad way

robbavey:
I've taken out `timecop` as it just confuses the situation even further.

andsel:
LGTM

Issue #30, opened 2020-05-05
make input plugin compatible with sequel fallback possibility from drivermanager to java connect


111andre111:
For JDBC input plugin (not sure about the other plugins) it seems like for the moment the sequel fallback possibility doesn't work, because there are some JDBC libraries out there that don't autoload or don't work through Drivermanager, but work with normal JDBC connect.
For that exists :driver parameter as an option which can be transferred as class to sequel and the fallback will work.
https://github.com/jeremyevans/sequel/blob/master/doc/opening_databases.rdoc#jdbc--

Here a small test snippet tested with irb:
```
require 'JDBC.jar'
require 'other.jar'
require "java"
require "sequel"
require "sequel/adapters/jdbc"
DBClass=Sequel::JDBC.load_driver('com......jdbc.Driver')
opts = { :user => "", :password => "", :pool_timeout => 5, :keep_reference => false, :driver => DBClass }
url = 'jdbc:..........'
DB = Sequel.connect(url, opts=opts)
DB['SELECT ...'].to_a.inspect
exit()
```
So the important part is this:
`DBClass=Sequel::JDBC.load_driver('com......jdbc.Driver')`
followed by that
`:driver => DBClass`

The relevant part for that in sequel is here:
https://github.com/jeremyevans/sequel/blob/5b5c884d36d40b7aac62bc03bbb1db08795b6297/lib/sequel/adapters/jdbc.rb#L215-L227

Here is still the irb part without Sequel in order to show what happens:
```
require 'JDBC.jar'
require 'other.jar'
require "java"
require "sequel"
require "sequel/adapters/jdbc"
driver = Java::com.....Driver.new
props = java.util.Properties.new
props.setProperty("user", "")
props.setProperty("password", "")
url = 'jdbc:....'
# now drivermanager test
conn = java.sql.DriverManager.get_connection(url)
# Error message No suitable driver found for jdbc:...
# Now try to fallback
conn = driver.connect(url, props)
# => #<Java::Com loaded
stmt = conn.createStatement();
sql = "SELECT count(*) AS COUNT FROM logstashtable";
rs = stmt.executeQuery(sql);
while (rs.next) do
  countname = rs.getString("COUNT");
  puts countname
end
stmt.close
conn.close
```

Now trying to translate that into logstash-input-jdbc code the place where something similar happens is here:
[...]

Issue #23, opened 2020-03-20
SyBase driver is not recognized by JDBC input


andsel:
In a `jdbc_input` configuration that uses SyBase driver (jConnect4) where the driver is loaded with setting `jdbc_driver_library` the plugin is not able to connect failing with:
```
[2020-03-20T17:35:37,348][ERROR][logstash.inputs.jdbc     ][main][970e2af7d6ed7910295d5e85805979ad1f2911d0369784f350e07fd298367f5a] Unable to connect to database. Tried 1 times {:error_message=>"Java::JavaSql::SQLException: No suitable driver found for jdbc:sybase:Tds:127.0.0.1:5000/testdb"}
[2020-03-20T17:35:37,374][ERROR][logstash.javapipeline    ][main][970e2af7d6ed7910295d5e85805979ad1f2911d0369784f350e07fd298367f5a] A plugin had an unrecoverable error. Will restart this plugin.
  Pipeline_id:main
  Plugin: <LogStash::Inputs::Jdbc jdbc_user=>"sa", use_column_value=>true, tracking_column=>"item_id", jdbc_password=><password>, statement=>"SELECT * FROM dbo.items > :sql_last_value", jdbc_driver_library=>"/home/andrea/.local/share/DBeaverData/drivers/drivers/sybase/jconnect/jconn4.jar", jdbc_connection_string=>"jdbc:sybase:Tds:127.0.0.1:5000/testdb", id=>"970e2af7d6ed7910295d5e85805979ad1f2911d0369784f350e07fd298367f5a", jdbc_driver_class=>"com.sybase.jdbc4.jdbc.SybDriver", enable_metric=>true, codec=><LogStash::Codecs::Plain id=>"plain_17de0ee4-e7fa-48fd-9c94-6c92325a201e", enable_metric=>true, charset=>"UTF-8">, jdbc_paging_enabled=>false, jdbc_page_size=>100000, jdbc_validate_connection=>false, jdbc_validation_timeout=>3600, jdbc_pool_timeout=>5, sql_log_level=>"info", connection_retry_attempts=>1, connection_retry_attempts_wait_time=>0.5, plugin_timezone=>"utc", last_run_metadata_path=>"/home/andrea/.logstash_jdbc_last_run", tracking_column_type=>"numeric", clean_run=>false, record_last_run=>true, lowercase_column_names=>true, use_prepared_statements=>false>
[...]

PR #34, opened 2020-06-08
driver loading: provide loaded driver to Sequel::connect

yaauie:
`Sequel::connect` will first attempt to auto-detect the driver based on the
jdbc connection URI, and will [fall-back][] to the driver given as `opts[:driver]`
when auto-detection fails.

By memoizing the loaded driver from our previous `Sequel::JDBC::load_driver`
and providing it to `Sequel::connect`, we provide one more chance for loading
to work.

This specifically seems to be the only path forward for Sybase drivers, whose
jdbc connection strings begin with `jdbc:sybase:Tds:` but use the driver with
path `com.sybase.jdbc4.jdbc.SybDriver`.

[fall-back]: https://github.com/jeremyevans/sequel/blob/5.19.0/lib/sequel/adapters/jdbc.rb#L211-L213

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.0.4
+  - Fixed issue where JDBC Drivers that don't correctly register with Java's DriverManager fail to load (such as Sybase) [#34](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/34)
+
 ## 5.0.3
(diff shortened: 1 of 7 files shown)


yaauie:
Fixes: https://github.com/logstash-plugins/logstash-integration-jdbc/issues/30
Fixes: https://github.com/logstash-plugins/logstash-integration-jdbc/issues/23

PR #37, opened 2020-06-16
Fix: properly pass (symbolized) sequel opts

kares:
in case of the **jdbc input** `sequel_opts => ...` were being passed with string keys (the streaming plugin symbolizes keys).
this is fine e.g. for passing `sequel_opts => { "jdbc_properties" => { ... } }` as sequel handles the `"jdbc_properties"` case but officially it's documented as `:jdbc_properties`. 

string keys turned out problematic when trying to disable quoting using: `sequel_opts => { quote_identifiers => false }` the option wasn't being picked from the Hash, as Sequel only checks for `:quote_identifiers` key.

also took this opportunity to introduce some sharing between jdbc and jdbc_streaming plugins as I noticed some of the previously fixed issues (e.g. driver loading) could also affect the (lesser used) jdbc_streaming plugin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.0.5
+  - Fixed user sequel_opts not being passed down properly [#37](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/37)
+  - Refactored jdbc_streaming to share driver loading, so the fixes from the jdbc plugin also effect jdbc_streaming
+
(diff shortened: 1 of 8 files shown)


kares:
Thanks Rob, just realized this isn't enough for me to be able to set `sequel_opts => { quote_identifiers => false }`
As `false` value will come back as `"false"` (String) and thus considered truthy by Sequel, so need to manually handle booleans.

PR #40, opened 2020-07-20
Doc:Replace plugin_header file with plugin_header-integration file

karenzone:
Integration plugins need a header that is different from the stand-alone plugin
header. The plugin docs links should point to the integration repo rather than the
stand-alone plugin repos. The header also adds a note that the plugin is part of
an integration rather than stand-alone.

The new header included in this work was added with elastic/logstash#11891 and elastic/logstash#12163 for the Logstash Reference, and elastic/logstash-docs#891 for the Versioned Plugin Reference.


jsvd (approved):
LGTM

karenzone:
**PUBLISHED:**
Screen Shot 2020-08-20 at 11 05 37 AM

PR #49, opened 2020-09-30
Doc:Highlight difference between static and streaming filter

karenzone:
Update docs to help guide users to the best jdbc filter for their use case


karenzone:
@karenzone: [Note to self] The filter table description for jdbc_static and jdbc_streaming need to be updated with more substance

karenzone:
@jsvd I set up the structure and hit some obvious points. I would appreciate a recap of the situations users find themselves in when they use the wrong filter for their use case.

PR #59, opened 2021-03-05
Extracted template method for static filter

### LINKED ISSUE #60: Remove duplicated methods in LogStash::Filters::Jdbc::Lookup
In class `LogStash::Filters::Jdbc::Lookup` the methods 
https://github.com/logstash-plugins/logstash-integration-jdbc/blob/fda99158e1578244c213e220f3efb35267f3b326/lib/logstash/filters/jdbc/lookup.rb#L131-L156

and 

https://github.com/logstash-plugins/logstash-integration-jdbc/blob/fda99158e1578244c213e220f3efb35267f3b326/lib/logstash/filters/jdbc/lookup.rb#L158-L183

shares a lot of logic.
Extract common part to avoid code duplication.

andsel:
In class LogStash::Filters::Jdbc::Lookup the methods [fetch(local, event](https://github.com/logstash-plugins/logstash-integration-jdbc/blob/fda99158e1578244c213e220f3efb35267f3b326/lib/logstash/filters/jdbc/lookup.rb#L131-L156)), [call_prepared(local, event)](https://github.com/logstash-plugins/logstash-integration-jdbc/blob/master/lib/logstash/filters/jdbc/lookup.rb#L158-L183) shared a lot of logic.
Extracted common part in template method name 'retrieve_local_data(local, event, load_method_ref)' and separated
the differiation logic in methods:
- `load_data_from_prepared`, used to load data from prepared statement rows
- `load_data_from_local`, used to load data from normal statement rows

Fixes #60

### DIFF
--- a/lib/logstash/filters/jdbc/lookup.rb
+++ b/lib/logstash/filters/jdbc/lookup.rb
@@ -82,10 +82,11 @@ module LogStash module Filters module Jdbc
 
     def enhance(local, event)
       if @prepared_statement
-        result = call_prepared(local, event)
+        load_method_ref = method(:load_data_from_prepared)
       else
-        result = fetch(local, event) # should return a LookupResult
+        load_method_ref = method(:load_data_from_local)
       end
+      result = retrieve_local_data(local, event, &load_method_ref) # should return a LookupResult
       if result.failed? || result.parameters_invalid?
         tag_failure(event)
       end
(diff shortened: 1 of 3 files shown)


lib/logstash/filters/jdbc/lookup.rb
```diff
@@ -82,10 +82,11 @@ def formatted_errors
 
     def enhance(local, event)
       if @prepared_statement
-        result = call_prepared(local, event)
+        load_method_ref = method(:load_data_from_prepared)
       else
-        result = fetch(local, event) # should return a LookupResult
+        load_method_ref = method(:load_data_from_local)
       end
```
kares, line 88: could we move this whole block to happen during `LookupProcessor#initialize`?
`enhance` is called on `plugin.filter` so we end up doing the method lookup every time.
we should be able to keep the looked up method in an instance field as `@prepared_statement` is known ahead.

kares (commented):
good refactoring.
feel like we should take it a bit further to avoid method lookup on `plugin.filter`

kares (approved):
:+1: nice refactoring

PR #62, opened 2021-04-15
Feat: try hard to log Java cause (chain)

kares:
as an attempt to resolve: https://github.com/logstash-plugins/logstash-integration-jdbc/issues/63

NOTE: kind of tested already: https://travis-ci.com/github/logstash-plugins/logstash-integration-jdbc/jobs/498812730#L600


andsel (approved):
This approach is patchy but serves the scope.
LGTM

PR #64, opened 2021-04-27
fix multiple JDBC libraries bug

marcel-richter:
I have tried to write a unit test but failed because of my poor ruby knowledge.
I guess testing this particular line is not total straight forward and requires some kind of mock.


marcel-richter:
I have signed the contributor agreement several times. The check doesn't get green. Don't know which magic trick this requires...

PR #69, opened 2021-05-18
Add `target` option to jdbc input

### LINKED ISSUE #19: Implement ECS-Compatibility Mode
_This is a stub issue, and needs to be fleshed out with details specific to
this plugin._

***

As a part of the effort to make plugins able to run in an ECS-Compatible manner
by default in an upcoming release of Logstash, this plugin needs to either
implement an ECS-Compatibility mode or certify that it does not implicitly use
fields that conflict with ECS.

----

Introduce a `target` option that specifies the name of the sub-field where nest the row columns loaded from DB. This option should be populated if ECS support is `not disabled`.

andsel:
This PR adds the `target` option to configure destination field inside the event to the JDBC input.
When ECS is enabled and the `target` is not specified it log a warn message.

Related #19

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -221,7 +221,6 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
   public
 
   def register
-    puts "register invoked"
     @logger = self.logger
     require "rufus/scheduler"
     prepare_jdbc_connection
@@ -271,7 +270,6 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
 
     # target must be populated if ecs_compatibility is not :disabled
     if @target.nil? && ecs_compatibility != :disabled
-      puts "@target is nil and ecs_compatibility: #{ecs_compatibility}"
       logger.warn("When ECS compatibility is enabled also target option must be valued")
     end
   end # def register
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -4,6 +4,7 @@ require "logstash/namespace"
 require "logstash/plugin_mixins/jdbc/common"
 require "logstash/plugin_mixins/jdbc/jdbc"
 require "logstash/plugin_mixins/ecs_compatibility_support"
+require "logstash/plugin_mixins/validator_support/field_reference_validation_adapter"
 
 # this require_relative returns early unless the JRuby version is between 9.2.0.0 and 9.2.8.0
 require_relative "tzinfo_jruby_patch"
@@ -132,6 +133,8 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
   include LogStash::PluginMixins::Jdbc::Jdbc
   # adds ecs_compatibility config which could be :disabled or :v1
   include LogStash::PluginMixins::ECSCompatibilitySupport(:disabled,:v1)
+  # adds :field_reference validator adapter
+  extend LogStash::PluginMixins::ValidatorSupport::FieldReferenceValidationAdapter
 
   config_name "jdbc"
 
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -274,7 +274,7 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
     # target must be populated if ecs_compatibility is not :disabled
     if @target.nil? && ecs_compatibility != :disabled
       logger.info('ECS compatibility is enabled but no ``target`` option was specified, it is recommended'\
-                  ' to set the option to avoid potential schema conflicts (if you\'re data is ECS compliant or'\
+                  ' to set the option to avoid potential schema conflicts (if your data is ECS compliant or'\
                   ' non-conflicting feel free to ignore this message)')
     end
   end # def register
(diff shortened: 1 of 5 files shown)


lib/logstash/inputs/jdbc.rb
```diff
@@ -260,6 +270,11 @@ def register
         converters[encoding] = converter
       end
     end
+
+    # target must be populated if ecs_compatibility is not :disabled
+    if @target.nil? && ecs_compatibility != :disabled
+      logger.warn("When ECS compatibility is enabled also target option must be valued")
```
kares, line 276: believe we agreed upon this being logger at the `info` level (as a compromise since it might be legit to skip setting `target` in cases where data has no conflicts with the schema).

what I would be also interested, due other plugins, is a standard message we would use across plugins ... draft:
`ECS compatibility is enabled but no ``target`` option was specified, it is recommended to set the option to avoid potential schema conflicts (if you're data is ECS compliant or non-conflicting feel free to ignore this message)` // cc @yaauie @karenzone

kares (commented):
:+1: code-wise looking good, 
we'll need some input on "standardizing" the missing target message

lib/logstash/inputs/jdbc.rb
andsel: +1 on standardize, maybe we can put it as constant in the ECS mixin?

logstash-integration-jdbc.gemspec
```diff
@@ -36,6 +36,8 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'tzinfo-data'
   # 3.5 limitation is required for jdbc-static loading schedule
   s.add_runtime_dependency 'rufus-scheduler', '< 3.5'
+  s.add_runtime_dependency 'logstash-mixin-ecs_compatibility_support', '~>1.1'
```
kaisecheng, line 39: the latest version is `1.2` which adds support for Elastic stack 8, ECS v8. Potentially, we may need to come back and add ECS v8 to all plugins that use `1.1` before releasing Logstash 8. I would recommend using the latest version now.
`1.2` [advance usage](https://github.com/logstash-plugins/logstash-mixin-ecs_compatibility_support#usage-advanced), [source](https://github.com/logstash-plugins/logstash-mixin-ecs_compatibility_support/pull/5)

andsel: good point

lib/logstash/inputs/jdbc.rb
```diff
@@ -129,6 +131,11 @@
 module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
   include LogStash::PluginMixins::Jdbc::Common
   include LogStash::PluginMixins::Jdbc::Jdbc
+  # adds ecs_compatibility config which could be :disabled or :v1
+  include LogStash::PluginMixins::ECSCompatibilitySupport(:disabled,:v1)
```
kaisecheng, line 135:
```suggestion
  include LogStash::PluginMixins::ECSCompatibilitySupport(:disabled,:v1,:v8 => :v1)
```

lib/logstash/inputs/jdbc.rb
yaauie: @kares In another conversation for a different plugin, I believe we discussed emitting this as `info`, since it isn't exactly something that needs to be resolved.

andsel: Can you recall maybe the plugin?So that I can use the same log info, to be consistent across various plugins

kares: > believe we agreed upon this being logger at the `info` level 

yes we agreed on info level logging previously, we do not have another plugin with the "warning" yet.
that's why I proposed a message for review, if accepted we would use it here and maybe move to ECS support mixing for re-use? (it's going to concern 3 other plugins I can think of right away)

kares:
```suggestion
      logger.info("ECS compatibility is enabled but no ``target`` option was specified, it is recommended to set the option to avoid potential schema conflicts (if you're data is ECS compliant or non-conflicting feel free to ignore this message)")
```

andsel: I think your message is perfect, in case of jdbc static and filter we would also add the `id` of the interested lookups, so in that case it could hit many lines like this, one for each lookup definition

lib/logstash/inputs/jdbc.rb
```diff
@@ -273,7 +273,9 @@ def register
 
     # target must be populated if ecs_compatibility is not :disabled
     if @target.nil? && ecs_compatibility != :disabled
-      logger.warn("When ECS compatibility is enabled also target option must be valued")
+      logger.info('ECS compatibility is enabled but no ``target`` option was specified, it is recommended'\
+                  ' to set the option to avoid potential schema conflicts (if you\'re data is ECS compliant or'\
```
kares, line 277:
```suggestion
                  ' to set the option to avoid potential schema conflicts (if your data is ECS compliant or'\
```

kares (approved):
just a slight typo that I had in the suggestion

spec/inputs/jdbc_spec.rb
```diff
@@ -328,6 +328,50 @@
...
+  context "when using target option is not set and ecs_compatibility is enabled" do
+    let(:settings) do
+      {
+        "statement" => "SELECT * from test_table FETCH FIRST 1 ROWS ONLY",
+        "ecs_compatibility" => :v1
+      }
+    end
+
+    it "should log a warn of missed target usage" do
+      expect(plugin.logger).to receive(:info).once.with('ECS compatibility is enabled but no ``target``'\
```
kares, line 367:
```suggestion
      expect(plugin.logger).to receive(:info).once.with(/ECS compatibility is enabled but no .*?target.*?/)
```

PR #71, opened 2021-05-24
Ensure `target` option is defined in Lookups when ECS is enabled

andsel:
This PR add the validation of `target` sub-option in the blocks of `local_lookups` for `logstash-filter-jdbc_static`.

The `target` attribute already existed, in this case it's only added a check that it's valued in case ECS is enabled.

Related to PR #69
Related to issue #19

### DIFF
--- a/lib/logstash/filters/jdbc_static.rb
+++ b/lib/logstash/filters/jdbc_static.rb
@@ -218,7 +218,7 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
     options["lookup_jdbc_driver_class"] = @lookup_jdbc_driver_class
     options["lookup_jdbc_driver_library"] = @lookup_jdbc_driver_library
     options["lookup_jdbc_connection_string"] = @lookup_jdbc_connection_string
-    options["ecs_compatibility"] = @ecs_compatibility
+    options["ecs_compatibility"] = ecs_compatibility
     options
   end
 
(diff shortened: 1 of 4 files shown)


lib/logstash/filters/jdbc_static.rb
```diff
@@ -214,6 +218,7 @@ def global_lookup_options(options = Hash.new)
     options["lookup_jdbc_driver_class"] = @lookup_jdbc_driver_class
     options["lookup_jdbc_driver_library"] = @lookup_jdbc_driver_library
     options["lookup_jdbc_connection_string"] = @lookup_jdbc_connection_string
+    options["ecs_compatibility"] = @ecs_compatibility
```
kares, line 221: believe [we should not rely](https://github.com/logstash-plugins/logstash-mixin-ecs_compatibility_support/blob/v1.2.0/lib/logstash/plugin_mixins/ecs_compatibility_support.rb#L69-L71) on the raw value but instead use the accessor method : 
```suggestion
    options["ecs_compatibility"] = ecs_compatibility
```
... while this works with the LegacyAdapter LS' implementation does not have the same value for `@ecs_compatibility` and the (accessor) `ecs_compatibility` method

kares (changes requested):
looking sharp, one slight change related to not relying on the i-var

kares (approved):
LGTM

PR #56, opened 2020-11-30
Update filter-jdbc_static.asciidoc

tbennett6421:
see https://github.com/logstash-plugins/logstash-integration-jdbc/issues/55

Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/

PR #72, opened 2021-05-26
Bump to v5.1.1

karenzone:
Bumps version to publish and pick up changes in #56


karenzone:
@andsel It looks like we didn't publish at 5.1.0.  Is there any reason we shouldn't publish after this PR is approved?

andsel:
I didn't publish because I thought that all the work on ECS should land in one release, instead of have 2 close. So If you want we could bring also the PR #56 in the `5.1.0`. WDYT?

karenzone:
I'd be totally on board if it were my PR. I'm more hesitant to bundle it with everything else because it's from a contributor, and I don't want it to get lost amidst all the ECS work.  

We shouldn't skip publishing at a minor (5.1.0), so perhaps bundling is our best option.  

If you ever want to put up a PR but hold off publishing, we have the provision to use `## Next` in the changelog description.

karenzone:
@andsel  Nice solution!  I'll follow up with the merge/publish for this one after your done publishing.

andsel (approved):
LGTM

karenzone:
Published at v5.1.1: https://rubygems.org/gems/logstash-integration-jdbc/versions/5.1.1

PR #76, opened 2021-06-23
Added spec to verify extract_from with nested field

### LINKED ISSUE #75: prepared_statement_bind_values in jdbc_streaming doesn't handle nested fields
This is because the ParameterHandler class tries to identify if there's either an interpolated reference (i.e. "%{[field]}") or a field reference (e.g. "[field]") in the values of prepared_statement_bind_values.

However, to detect the field references, [a regex is used](https://github.com/logstash-plugins/logstash-integration-jdbc/blob/master/lib/logstash/plugin_mixins/jdbc_streaming/parameter_handler.rb#L29-L31):

```ruby
if given_value =~ /\A\s*\[[^\]]+\]\s*\z/
  return FieldParameter.new(given_value)
end
```

This regex can't handle nested field references:

```
irb(main):003:0> "[field]" =~ /\A\s*\[[^\]]+\]\s*\z/
=> 0
irb(main):004:0> "[nested][field]" =~ /\A\s*\[[^\]]+\]\s*\z/
=> nil
```

For now the workaround is to replace a `"[nested][field]"` reference with a `"%{[nested][field]}"` interpolation

andsel:
Fixes #75

### DIFF
--- a/spec/plugin_mixins/jdbc_streaming/parameter_handler_spec.rb
+++ b/spec/plugin_mixins/jdbc_streaming/parameter_handler_spec.rb
@@ -1,0 +1,23 @@
+# encoding: utf-8
+require "logstash/devutils/rspec/spec_helper"
+require "logstash/plugin_mixins/jdbc_streaming/parameter_handler"
+
+
+describe LogStash::PluginMixins::JdbcStreaming::ParameterHandler do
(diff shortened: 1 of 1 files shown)


jsvd (approved):
LGTM thank you for taking care of this issue @andsel ❤️

PR #82, opened 2021-07-28
ECS: refinements

yaauie:
- use shared Target Check
 - use EventFactory to create events


kares (approved):
:icecream: - should be the last tier 1 with missing `event_factory` support

*could use a changelog but otherwise lgtm!*

PR #85, opened 2021-10-26
Fix connection lock to unblock pipeline reload

### LINKED ISSUE #80: jdbc input errors at start-up prevent pipeline termination/reloading
Connectivity issues in during startup of jdbc input such as an invalid password are retried forever.
During these retries the status of `stop?` is never checked so the pipeline blocks reloadings.

The issue comes from the rufus scheduler by default catching exceptions in the jobs it executes. It's necessary to catch these exceptions and terminate the scheduler if the pipeline is terminating.

kaisecheng:
## Release notes

shutdown gracefully and unblock pipeline reload when connectivity issues happen in jdbc input such as an invalid password

## What does this PR do?

This commit moves the connection lock to `begin`, `end` block to ensure it is unlocked in case of exception.
Prior to this change, exception throw to rufus-scheduler printing a page of rufus-scheduler error message. Now it is caught and log the necessary part.

Logstash fail to reload the pipeline from central pipeline management if the jdbc input config with wrong password.
The reload thread waits [ReentrantLock](https://github.com/logstash-plugins/logstash-integration-jdbc/blob/v5.1.6/lib/logstash/plugin_mixins/jdbc/jdbc.rb#L199) that is held by the [worker thread](https://github.com/logstash-plugins/logstash-integration-jdbc/blob/v5.1.6/lib/logstash/plugin_mixins/jdbc/jdbc.rb#L211)
#80

## Checklist

- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~

## Author's Checklist

- [ ]

Start Logstash with wrong postgres password
```
    jdbc {
        jdbc_driver_library => "/PATH/TO/POSTGRES/postgresql-42.2.14.jar"
        jdbc_driver_class => "org.postgresql.Driver"
        jdbc_connection_string => "jdbc:postgresql://localhost:5432/postgres"
        jdbc_user => "postgres"
        jdbc_password => "WRONG_PASSWORD"
        schedule => "* * * * *"
        statement => "select * from pg_catalog.pg_am"
    }
```
## Related issues

- closes #80


andsel:
HI @kaisecheng with PR https://github.com/elastic/logstash-devutils/pull/99 I've moved the "Logstash timstamp equivalent" matcher to `devutils 2.3.0` so you can drop from this PR and reuse it from devutils. Thanks to highlight it so that we moved in the shared gem.
After use it, please put a check mark in issue https://github.com/elastic/logstash-devutils/issues/98 so that we can close it

PR #90, opened 2021-11-07
Add another flexible way for JDBC paging

dingx1:
## Release notes

Add configuration option to avoid the first count statement in paginated queries.

## What does this PR do?

This PR introduce a new configuration option to avoid the initial count select statement executed by `Sequel` in case of paginated queries. The paginated queries could be done now with initial count or not, in case `jdbc_paging_avoid_count` is enabled the plugin executes paged queries till it reach a page with less rows than the expected, instead to rely on the total row count.

In some circumstances, like stored procedure call with pagination bounds, the initial count query is simply not meaningful.
In cases where the pages to retrieve are few the initial count query could be an useful overhead.

## Checklist

- [x] I have made corresponding changes to the documentation

## Author's Checklist

- [x] check it with a local Logstash test.

- spin up a local DB
- configure a JDBC input query with pagination and with the new `jdbc_paging_avoid_count` and check it retrieves all the expected rows. In logs should be present a line for each paged query

```
input {
  jdbc {
    jdbc_driver_library => "/path/mysql-connector-java-8.0.26.jar"
    jdbc_driver_class => "Java::com.mysql.cj.jdbc.Driver"
    jdbc_connection_string => "jdbc:mysql://localhost:3306/test_logstash"

    jdbc_user => "user"
    jdbc_password => "s3cret"

    jdbc_default_timezone => "UTC"

    schedule => "* * * * *"

    use_column_value => true
    tracking_column => "log_id"
    tracking_column_type => "numeric"
    last_run_metadata_path => ".last_run"
    
    statement => "SELECT * FROM data_log WHERE log_id > :sql_last_value"
    jdbc_paging_enabled => true
    jdbc_page_size => 7
    jdbc_paging_avoid_count => true
  }
}

output {
  stdout {
    codec => "rubydebug"
  }
}
```

## Related issues

-

### DIFF
--- a/docs/input-jdbc.asciidoc
+++ b/docs/input-jdbc.asciidoc
@@ -385,12 +385,12 @@ Whether to use count query during the JDBC paging
 If false, a count query followed by multiple paged queries (with `LIMIT` statement, etc.) are used under the hood.
 
 If true, multiple paged queries (without a count query ahead) will be performed until no more rows are retrieved.
-The `offset` and `size` parameters can be used in the statement (`size` equal to `jdbc_page_size`, and `offset` incremented by `size`).
+The `offset` and `size` parameters must be used in your statement (`size` equal to `jdbc_page_size`, and `offset` incremented by `size`).
 Example:
 
 [source, ruby]
 ------------------------------------------------------
-"SELECT id, mycolumn1, mycolumn2 FROM my_table WHERE id > :sql_last_value LIMIT :offset, :size"
+"SELECT id, mycolumn1, mycolumn2 FROM my_table WHERE id > :sql_last_value LIMIT :size OFFSET :offset"
 ------------------------------------------------------
 
 [source, ruby]
--- a/docs/input-jdbc.asciidoc
+++ b/docs/input-jdbc.asciidoc
@@ -153,7 +153,7 @@ NOTE: Not all JDBC accessible technologies will support prepared statements.
 With the introduction of Prepared Statement support comes a different code execution path and some new settings. Most of the existing settings are still useful but there are several new settings for Prepared Statements to read up on.
 Use the boolean setting `use_prepared_statements` to enable this execution mode. Use the `prepared_statement_name` setting to specify a name for the Prepared Statement, this identifies the prepared statement locally and remotely and it should be unique in your config and on the database. Use the `prepared_statement_bind_values` array setting to specify the bind values, use the exact string `:sql_last_value` (multiple times if necessary) for the predefined parameter mentioned before. The `statement` (or `statement_path`) setting still holds the SQL statement but to use bind variables you must use the `?` character as a placeholder in the exact order found in the `prepared_statement_bind_values` array.
 
-NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries (when `jdbc_page_size` is greater than `0`) under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
+NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries (when `jdbc_paging_avoid_count` is false) under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
 
 Example:
 [source,ruby]
@@ -193,6 +193,7 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 | <<plugins-{type}s-{plugin}-jdbc_fetch_size>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-jdbc_page_size>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-jdbc_paging_enabled>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-jdbc_paging_avoid_count>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-jdbc_password>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-jdbc_password_filepath>> |a valid filesystem path|No
 | <<plugins-{type}s-{plugin}-jdbc_pool_timeout>> |<<number,number>>|No
@@ -359,22 +360,6 @@ JDBC fetch size. if not provided, respective driver's default will be used
 
 JDBC page size
 
-If greater than `0`, a count query and multiple paged queries (with `LIMIT` statement, etc.) are used under the hood.
-
-Otherwise, if less than `0`, multiple queries (without count query) will be performed until no rows are returned.
-The `:offset` and `:size` parameters can be used in the statement (`:size` equal to `-jdbc_page_size`, and `:offset` incremented by `:size`).
-Example:
-
-[source, ruby]
------------------------------------------------
-"SELECT id, mycolumn1, mycolumn2 FROM my_table WHERE id > :sql_last_value LIMIT :offset, :size"
------------------------------------------------
-
-[source, ruby]
------------------------------------------------
-"CALL fetch_my_data(:sql_last_value, :offset, :size)"
------------------------------------------------
-
 [id="plugins-{type}s-{plugin}-jdbc_paging_enabled"]
 ===== `jdbc_paging_enabled`
 
@@ -388,6 +373,30 @@ Each query will use limits and offsets to collectively retrieve the full
 result-set. The limit size is set with `jdbc_page_size`.
 
 Be aware that ordering is not guaranteed between queries.
+
+[id="plugins-{type}s-{plugin}-jdbc_paging_avoid_count"]
+===== `jdbc_paging_avoid_count`
+
+  * Value type is <<boolean,boolean>>
+  * Default value is `false`
+
+Whether to use count query during the JDBC paging
+
+If false, a count query and multiple paged queries (with `LIMIT` statement, etc.) are used under the hood.
+
+Otherwise, if true, multiple paged queries (without a count query ahead) will be performed until no more rows are retrieved.
+The `offset` and `size` parameters can be used in the statement (`size` equal to `jdbc_page_size`, and `offset` incremented by `size`).
+Example:
+
+[source, ruby]
+------------------------------------------------------
+"SELECT id, mycolumn1, mycolumn2 FROM my_table WHERE id > :sql_last_value LIMIT :offset, :size"
+------------------------------------------------------
+
+[source, ruby]
+------------------------------------------------------
+"CALL fetch_my_data(:sql_last_value, :offset, :size)"
+------------------------------------------------------
 
 [id="plugins-{type}s-{plugin}-jdbc_password"]
 ===== `jdbc_password`
--- a/docs/input-jdbc.asciidoc
+++ b/docs/input-jdbc.asciidoc
@@ -153,7 +153,7 @@ NOTE: Not all JDBC accessible technologies will support prepared statements.
 With the introduction of Prepared Statement support comes a different code execution path and some new settings. Most of the existing settings are still useful but there are several new settings for Prepared Statements to read up on.
 Use the boolean setting `use_prepared_statements` to enable this execution mode. Use the `prepared_statement_name` setting to specify a name for the Prepared Statement, this identifies the prepared statement locally and remotely and it should be unique in your config and on the database. Use the `prepared_statement_bind_values` array setting to specify the bind values, use the exact string `:sql_last_value` (multiple times if necessary) for the predefined parameter mentioned before. The `statement` (or `statement_path`) setting still holds the SQL statement but to use bind variables you must use the `?` character as a placeholder in the exact order found in the `prepared_statement_bind_values` array.
 
-NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries (when `jdbc_paging_avoid_count` is false) under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
+NOTE: Building count queries around a prepared statement is not supported at this time. Because jdbc paging uses count queries when `jdbc_paging_manual_mode` is false，jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
 
 Example:
 [source,ruby]
@@ -193,7 +193,7 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 | <<plugins-{type}s-{plugin}-jdbc_fetch_size>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-jdbc_page_size>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-jdbc_paging_enabled>> |<<boolean,boolean>>|No
-| <<plugins-{type}s-{plugin}-jdbc_paging_avoid_count>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-jdbc_paging_manual_mode>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-jdbc_password>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-jdbc_password_filepath>> |a valid filesystem path|No
 | <<plugins-{type}s-{plugin}-jdbc_pool_timeout>> |<<number,number>>|No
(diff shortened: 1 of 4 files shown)


andsel:
Hi @dingx1 thank you for your contribution to the project!
Please could you describe better how your PR avoid the usage of count in paged queries, please or provide an integration test that verifies the changes

dingx1:
Hi @andsel sorry for the lack of detailed description.

Assume that the database is mysql, and there are 50 rows in `test_table`.

### **Existing paging mode**
Use the following configuration:
```
{
  "statement" => "SELECT * from test_table",
  "jdbc_paging_enabled" => true,
  "jdbc_page_size" => 20
}
```
The following statements will be executed:
1. SELECT count(*) AS `count` FROM (SELECT * FROM test_table) AS `t1` LIMIT 1
1. SELECT * FROM (SELECT * FROM test_table) AS `t1` LIMIT 20 OFFSET 0
1. SELECT * FROM (SELECT * FROM test_table) AS `t1` LIMIT 20 OFFSET 20
1. SELECT * FROM (SELECT * FROM test_table) AS `t1` LIMIT 20 OFFSET 40

Issues:
1. Additional subqueries may increase performance overhead.
2. Unfriendly to stored procedures.

### **Paging mode in this pr**
Use the following configuration:  
(jdbc_page_size is a negative value):
```
{
  "statement" => "SELECT * from test_table LIMIT :size OFFSET :offset",
  "jdbc_paging_enabled" => true,
  "jdbc_page_size" => -20
}
```
The following statements will be executed:
1. SELECT * FROM test_table LIMIT 20 OFFSET 0
1. SELECT * FROM test_table LIMIT 20 OFFSET 20
1. SELECT * FROM test_table LIMIT 20 OFFSET 40

For the third query, Only 10 rows (less than 20) are retrieved. Therefore, no more queries need to be performed, and the jdbc paging ended.

andsel:
Thanks @dingx1 for clarification, now I understand it better.

I've some thoughts on this, it avoids to execute the initial count query, which is usually a fast query on many database, so I think that the cost of it is absorbed by the subsequent paged queries. It could provide some benefits in the cases where the number of records to load fits in a single page.

I think that the use of negative values to enable a behavior feels like a cheat code, probably is better to add a new configuration option, name it `jdbc_paging_avoid_count` or something similar, that could be enabled if and only if the `jdbc_paging_enabled` is defined. In this way it is more explicit the intention it provides.

We can steer this PR in that direction, WDYT?

dingx1:
Thanks @andsel for your idea, and I agree.
I have done so. Please check it.
Now, the configuration is as follow:

```
{
  "statement" => "SELECT * from test_table LIMIT :size OFFSET :offset",
  "jdbc_paging_enabled" => true,
  "jdbc_paging_avoid_count" => true,
  "jdbc_page_size" => 20
}
```

docs/input-jdbc.asciidoc
```diff
@@ -373,6 +374,30 @@ result-set. The limit size is set with `jdbc_page_size`.
...
 
+[id="plugins-{type}s-{plugin}-jdbc_paging_avoid_count"]
+===== `jdbc_paging_avoid_count`
+
+  * Value type is <<boolean,boolean>>
+  * Default value is `false`
+
+Whether to use count query during the JDBC paging
+
+If false, a count query and multiple paged queries (with `LIMIT` statement, etc.) are used under the hood.
```
andsel, line 385:
```suggestion
If false, a count query followed by multiple paged queries (with `LIMIT` statement, etc.) are used under the hood.
```

docs/input-jdbc.asciidoc
```diff
@@ -373,6 +374,30 @@ result-set. The limit size is set with `jdbc_page_size`.
...
+
+Otherwise, if true, multiple paged queries (without a count query ahead) will be performed until no more rows are retrieved.
```
andsel, line 387: Maybe is better the  "if true" so it's symmetric to the `false` case, a couple of eyes from @karenzone  on the documentation part are far better then mine on this.

```suggestion
If true, multiple paged queries (without a count query ahead) will be performed until no more rows are retrieved.
```

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -29,12 +29,25 @@ class NormalStatementHandler < StatementHandler
...
     # @yieldparam row [Hash{Symbol=>Object}]
-    def perform_query(db, sql_last_value, jdbc_paging_enabled, jdbc_page_size)
+    def perform_query(db, sql_last_value, jdbc_paging_enabled, jdbc_paging_avoid_count, jdbc_page_size)
       query = build_query(db, sql_last_value)
       if jdbc_paging_enabled
-        query.each_page(jdbc_page_size) do |paged_dataset|
-          paged_dataset.each do |row|
-            yield row
+        unless jdbc_paging_avoid_count
+          query.each_page(jdbc_page_size) do |paged_dataset|
```
andsel, line 36: I found 
```ruby
unless 
  #code segment A
else 
  #code segment B
end
```
a little bit counterintuitive

I would use the normal `if` and just swapping the code segments

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -29,12 +29,25 @@ class NormalStatementHandler < StatementHandler
...
+            paged_dataset.each do |row|
+              yield row
+            end
+          end
+        else
+          offset = 0
+          loop do
+            count = 0
```
andsel, line 44: Instead of count is better to use a more meaningful name that express the intent of the variable

```suggestion
            rows_in_page = 0
```

andsel (changes requested):
The code part is almost there, I left some nitpicks, what it's missed is the specification part. 
Unit and/or integration test would be the safe guards that this feature keeps working on future changes.
Would be great if you can add a simple test, maybe a table with 15 rows, page size of 10 and proof that the code change can load all the 15 rows.

For the doc part I would ask a look to @karenzone

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -29,12 +29,25 @@ class NormalStatementHandler < StatementHandler
...
       query = build_query(db, sql_last_value)
       if jdbc_paging_enabled
-        query.each_page(jdbc_page_size) do |paged_dataset|
-          paged_dataset.each do |row|
-            yield row
+        if jdbc_paging_avoid_count
+          offset = 0
+          loop do
+            rows_in_page = 0
+            query.with_sql(query.sql, offset: offset, size: jdbc_page_size).each do |row|
```
andsel, line 39: I tried locally and this line expected to be executed one time per page, but `offset` and `size` are not recognized as parameters, so it executes a full scan query and doesn't apply any pagination.
I inserted print statements before and after the query like in:
```ruby
puts "Executing query #{query.sql} offset: #{offset}, size: #{jdbc_page_size}"
query.with_sql(query.sql, offset: offset, size: jdbc_page_size).each do |row|
  yield row
  rows_in_page += 1
end
puts "rows_in_page: #{rows_in_page}"            
```

what I've obtained is:
```
Executing query SELECT * FROM data_log WHERE log_id > 0 offset: 0, size: 7
[2021-11-25T09:38:00,601][INFO ][logstash.inputs.jdbc     ][main][ff8e261c00f8975f017675a993b934c93e346dc2fe1ad66ed9c686980d264c2b] (0.013970s) SELECT * FROM data_log WHERE log_id > 0
rows_in_page: 39
```
while I would have expected the output of multiple lines in the form of:
```
[2021-11-25T09:38:00,601][INFO ][logstash.inputs.jdbc     ][main][ff8e261c00f8975f017675a993b934c93e346dc2fe1ad66ed9c686980d264c2b] (0.013970s) SELECT * FROM data_log WHERE log_id > 0 offset 0 rows 7

[2021-11-25T09:38:00,601+delta][INFO ][logstash.inputs.jdbc     ][main][ff8e261c00f8975f017675a993b934c93e346dc2fe1ad66ed9c686980d264c2b] (0.013970s) SELECT * FROM data_log WHERE log_id > 0 offset 7 rows 7;

[2021-11-25T09:38:00,601+delta][INFO ][logstash.inputs.jdbc     ][main][ff8e261c00f8975f017675a993b934c93e346dc2fe1ad66ed9c686980d264c2b] (0.013970s) SELECT * FROM data_log WHERE log_id > 0 offset 14 rows 7;
```

andsel (changes requested):
I've tried this inside Logstash running against a real DB (MySQL) and the pagination doesn't work.
I described the proof that it doesn't work, if you need a reproducer please ping me.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
dingx1: Hi, here's the thing.
When using this paging mode, the part of paging conditions has to be written by the user.
Therefore, the statement should be written as follow:
`
SELECT * FROM data_log WHERE log_id > :sql_last_value LIMIT :size OFFSET :offset
`
I found that it was also missing from my unit test code. I've corrected it.

andsel: Now I see it. My doubt is that for paged queries where `jdbc_paging_avoid_count` is `false` there is not need to remember to add paging statement in the query, `Sequel` does it. But when the `jdbc_paging_avoid_count` is `true` the user has to remember to add those. 
I think that to provide best experience, the query has to remain the same, so without intervention when switching the `jdbc_paging_avoid_count` on and off. It's to risky to let that in hands of a user, because he can forget to add those, and so inadvertently create maybe killing queries.

dingx1: Hi. Allow me to explain again from the beginning. It takes a little patience to read.

The goal of this PR is to deal with the following situations when `jdbc_paging_enabled` is `true`:

Assume that the database is mysql, and `jdbc_page_size` is `10`.

## 1. simple query
Assume the following statement:
```
SELECT * from test_table
```
The following statements will be created automatically by sequel:
* SELECT count(*) AS count FROM (SELECT * FROM test_table) AS t1 LIMIT 1
* SELECT * FROM (SELECT * FROM test_table) AS t1 LIMIT 10 OFFSET 0
* SELECT * FROM (SELECT * FROM test_table) AS t1 LIMIT 10 OFFSET 10
* subsequent ...

Issues:
1. When the amount of data is large, count query is unnecessary and expensive.
2. Wrapping user's statements with subqueries is also expensive on some database engine, such as mysql 5.6 and earlier version.
  
## 2. Complex SQL query
```
SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table LIMIT <offset>, <size>)
```
The following statements will be created automatically by sequel:
* SELECT count(*) AS count FROM ( SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table) ) AS t1 LIMIT 1
* SELECT * FROM ( SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table) ) AS t1 LIMIT 10 OFFSET 0
* SELECT * FROM ( SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table) ) AS t1 LIMIT 10 OFFSET 10
* subsequent ...

But, But what I expect is:
* SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table LIMIT 10 OFFSET 0)
* SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table LIMIT 10 OFFSET 10)
* subsequent ...

## 3. Stored procedure
```
CALL fetch_my_data(:sql_last_value, <offset>, <size>)
```
The following statements will be created automatically by sequel:
* SELECT count(*) AS count FROM ( CALL fetch_my_data(:sql_last_value) ) AS t1 LIMIT 1
* SELECT * FROM ( CALL fetch_my_data(:sql_last_value) ) AS t1 LIMIT 10 OFFSET 0
* SELECT * FROM ( CALL fetch_my_data(:sql_last_value) ) AS t1 LIMIT 10 OFFSET 10
* subsequent ...

Issues:
1. This syntax is not supported. At least MySQL is.
2. This is not what I want. What I expect is:
    * CALL fetch_my_data(:sql_last_value, 0, 10)
    * CALL fetch_my_data(:sql_last_value, 10, 10)

    I would decide how to page inside the stored procedure.

<br/>

---
Therefore, in order to solve the above problems, as a user, I really need a possibility of writing paging conditions myself.

Now, here's what I can do more:
1. Give `jdbc_paging_avoid_count` a new name that makes it more easier to understand its purpose, such as `jdbc_paging_manual_mode` / `jdbc_paging_custom_mode`. Or you give one.
2. Check the users' statement configuration and force the user to use the `:offset` and `:size` parameters in it.

And WDYT?

<br/>
The following are some related issues in old repository from other users, which also express similar demands, for reference.

https://hub.fastgit.org/logstash-plugins/logstash-input-jdbc/issues/345
https://hub.fastgit.org/logstash-plugins/logstash-input-jdbc/issues/332
https://hub.fastgit.org/logstash-plugins/logstash-input-jdbc/issues/307
https://hub.fastgit.org/logstash-plugins/logstash-input-jdbc/issues/305
https://hub.fastgit.org/logstash-plugins/logstash-input-jdbc/issues/298#issuecomment-416886991
https://hub.fastgit.org/logstash-plugins/logstash-input-jdbc/pull/387/files

docs/input-jdbc.asciidoc
```diff
@@ -153,7 +153,7 @@ NOTE: Not all JDBC accessible technologies will support prepared statements.
...
 
-NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
+NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries (when `jdbc_paging_avoid_count` is false) under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
```
karenzone, line 156:
```suggestion
NOTE: Building count queries around a prepared statement is not supported at this time. Because jdbc paging uses count queries when `jdbc_paging_avoid_count` is false, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
```

karenzone: Nice addition. Let's consider breaking this info into two sentences to make it easier to read and understand.

docs/input-jdbc.asciidoc
```diff
@@ -193,6 +193,7 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 | <<plugins-{type}s-{plugin}-jdbc_fetch_size>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-jdbc_page_size>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-jdbc_paging_enabled>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-jdbc_paging_avoid_count>> |<<boolean,boolean>>|No
```
karenzone, line 196: Please move the new entry above `jdbc_paging_enabled` to keep options in alpha order.

docs/input-jdbc.asciidoc
```diff
@@ -373,6 +374,30 @@ result-set. The limit size is set with `jdbc_page_size`.
...
+[source, ruby]
+------------------------------------------------------
+"SELECT id, mycolumn1, mycolumn2 FROM my_table WHERE id > :sql_last_value LIMIT :size OFFSET :offset"
+------------------------------------------------------
+
+[source, ruby]
+------------------------------------------------------
+"CALL fetch_my_data(:sql_last_value, :offset, :size)"
+------------------------------------------------------
+
```
karenzone, line 400: Please move the new option/info above `jdbc_paging_enabled` to keep options and sections in alpha order.

karenzone: Thanks for this info and the examples. The examples help.

karenzone (commented):
DOCS REVIEW: 
Nice job on the docs, @dingx1. Thanks for taking time to add such nice descriptions and the helpful examples. 

I left some minor suggestions inline for your consideration.  If you make any additional docs changes while finalizing this PR and would like another review, please ping me.  Otherwise, docs build cleanly, render as expected, and LGTM!

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
andsel: Thanks @dingx1 for the full explanation, now I understand it better.

I would prefer `jdbc_paging_manual_mode`, the intent seems more explicit.

We need more work on the documentation to expose this functionality more clean as we can, describing the use cases that it fits.

dingx1: Hi, I've renamed it and improved the documentation. Could you please review it again, thanks!

docs/input-jdbc.asciidoc
```diff
@@ -373,6 +374,52 @@ result-set. The limit size is set with `jdbc_page_size`.
...
+  * Default value is `false`
+
+Whether to use manual mode during the JDBC paging
+
+If false, your statement will be automatically surrounded by a count query and subsequent multiple paged queries (with `LIMIT` statement, etc.).
+
+If true, multiple queries (without a count query ahead) will be performed with your statement, until no more rows are retrieved.
+You have to write the paging conditions in your statement configuration by self.
+The `offset` and `size` parameters can be used in your statement (`size` equal to `jdbc_page_size`, and `offset` incremented by `size` for each query).
+When the number of rows returned by the query is not equal to 'size', JDBC paging will be ended.
```
andsel, line 390:
```suggestion
When the number of rows returned by the query is not equal to 'size', SQL paging will be ended.
```

docs/input-jdbc.asciidoc
```diff
@@ -373,6 +374,52 @@ result-set. The limit size is set with `jdbc_page_size`.
...
+
+  * Value type is <<boolean,boolean>>
+  * Default value is `false`
+
+Whether to use manual mode during the JDBC paging
+
+If false, your statement will be automatically surrounded by a count query and subsequent multiple paged queries (with `LIMIT` statement, etc.).
+
+If true, multiple queries (without a count query ahead) will be performed with your statement, until no more rows are retrieved.
+You have to write the paging conditions in your statement configuration by self.
```
andsel, line 388:
```suggestion
You have to write the paging conditions in your statement configuration by yourself.
```

andsel (changes requested):
LGTM left some minor notes on the docs part

docs/input-jdbc.asciidoc
karenzone: I would like to suggest another slightly different wording: 

"You have to write your own paging conditions in your statement configuration."

andsel (approved):
Thanks @dingx1 for all the effort the PR looks good to me.
I'll push a minor commit to bump the version, and then I'll publish it.

andsel:
Hi @dingx1 this PR is created from your local `master` and aligning it has some problems. I force pushed but lost you last commit. It's difficult to work on it for me, so please could you create a new PR branching from the `main` branch, name it for example `feature/query_manual_paging`. Sorry for the inconvenient :-(

andsel:
I created a support PR #94 to have a more clean workflow

andsel:
So you can create a new PR from commit https://github.com/logstash-plugins/logstash-integration-jdbc/commit/3b312fbde0c16dbb182b995b5ae5d8ae75a4c37a 
so:
```sh
git checkout 3b312fbde0c16dbb182b995b5ae5d8ae75a4c37a
git checkout -b feature/query_manual_paging
git push origin feature/query_manual_paging
```

In this way we should preserve all the work you have done and I can do a version bump

dingx1:
Hi @andsel I checked the commit logs. `3b312fb` is my final commit. You can use it directly, and append your revision after it. :-)

andsel:
@dingx1 
- the red CI on `8.x` is tracked by https://github.com/logstash-plugins/.ci/issues/31 and not related to your changes
- I close this PR because #95 exists
- on #95 I overwrite the description with that of this PR
- bump the new version on #95

PR #92, opened 2021-12-01
Ambiguous Timestamp Handling

yaauie:
Adds support for timezones provided to `jdbc_default_timezone` to include instructions for how to handle ambiguous times from daylight-savings related overlaps.

The SQL Timestamp column type does not contain offset information, so when it is used to store local times and those local times come from a timezone with seasonal daylight savings transitions, there is the possibility for a timestamp to be ambiguous.

When reading a timestamp from such a column that cannot be unambiguously resolved to a single point on an ordered and continuous timeline, a `TZInfo::AmbiguousTime` has been raised mid-processing, causing a `Sequel::InvalidValue` to be raised and subsequent rows in the result set to not be emitted.

This changeset allows a user to configure DST-related disambiguation when specifying their `jdbc_default_timezone`. When encountering an ambiguous timestamp in the 2 overlapping hours surrounding the transition from Daylight Savings time to Standard Time:

 - `jdbc_default_timezone => "America/Los_Angeles[dst_enabled_on_overlap:true]"`: ambiguous timestamps is assumed to be from _before_ the transition point
 - `jdbc_default_timezone => "America/Los_Angeles[dst_enabled_on_overlap:false]"`: ambiguous timestamps is assumed to be from _after_ the transition point
 - `jdbc_default_timezone => "America/Los_Angeles"`: ambiguous timestamps results in query failure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.2.0
+  - Added support for disambiguating timestamps in daylight-savings overlap periods
+
 ## 5.1.8
   - Fix the blocking pipeline reload and shutdown when connectivity issues happen [#85](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/85)
 
--- a/docs/input-jdbc.asciidoc
+++ b/docs/input-jdbc.asciidoc
@@ -282,6 +282,8 @@ JDBC connection string
 ===== `jdbc_default_timezone`
 
   * Value type is <<string,string>>
+  ** Value should be a canonical timezone or offset, such as `Europe/Paris` or `Etc/GMT+3`
+  ** Value _may_ include square-bracketed extensions, such as `America/Denver[dst_enabled_on_overlap:true]`
   * There is no default value for this setting.
 
 Timezone conversion.
@@ -295,7 +297,12 @@ in relative UTC time in ISO8601 format.
 
 Using this setting will manually assign a specified timezone offset, instead
 of using the timezone setting of the local machine.  You must use a canonical
-timezone, *America/Denver*, for example.
+timezone, `America/Denver`, for example.
+
+Ambiguous Timestamps.
+While it is common to store local times in SQL's timestamp column type, many timezones change their offset during the course of a calendar year and therefore cannot be used with SQL's timestamp type to represent an ordered, continuous timeline.
+For example in the `America/Chicago` zone when daylight savings ends in the autumn, the clock rolls from `01:59:59` back to `01:00:00`, making any timestamp in the 2-hour period between `01:00:00CDT` and `02:00:00CST` on that day ambiguous.
+When encountering an ambiguous timestamp caused by a DST transition, the query will fail unless the timezone specified here includes a square-bracketed instruction for how to handle overlapping periods (such as: `America/Chicago[dst_enabled_on_overlap:true]` or `Australia/Melbourne[dst_enabled_on_overlap:false]`).
 
 [id="plugins-{type}s-{plugin}-plugin_timezone"]
 ===== `plugin_timezone`
--- a/spec/inputs/jdbc_spec.rb
+++ b/spec/inputs/jdbc_spec.rb
@@ -1447,6 +1447,53 @@ describe LogStash::Inputs::Jdbc do
     end
   end
 
+  context "when retrieving records with ambiguous timestamps" do
+
+    # let(:value_tracker) { double("value tracker", :set_value => nil, :write => nil) }
+
+    let(:settings) do
+      {
+        "statement" => "SELECT * from types_table",
+        "jdbc_default_timezone" => jdbc_default_timezone
+      }
+    end
+
+    before(:each) do
+      db << "INSERT INTO types_table (num, string, started_at, custom_time, ranking) VALUES (1, 'A test', '1999-12-31', '2021-11-07 01:23:45', 95.67)"
+
+      plugin.register
+
+      # plugin.set_value_tracker(value_tracker)
+      # allow(value_tracker).to receive(:value).and_return("bar")
+    end
+
+    context "when initialized with a preference for DST being enabled" do
+      let(:jdbc_default_timezone) { 'America/Chicago[prefer-dst:true]' }
+
+      it 'treats the timestamp column as if DST was enabled' do
+        plugin.run(queue)
+        event = queue.pop
+        expect(event.get("custom_time")).to be_a_logstash_timestamp_equivalent_to("2021-11-07T06:23:45Z")
+      end
+    end
+    context "when initialized with a preference for DST being disabled" do
+      let(:jdbc_default_timezone) { 'America/Chicago[prefer-dst:false]' }
+
+      it 'treats the timestamp column as if DST was disabled' do
+        plugin.run(queue)
+        event = queue.pop
+        expect(event.get("custom_time")).to be_a_logstash_timestamp_equivalent_to("2021-11-07T07:23:45Z")
+      end
+    end
+    context "when initialized without a preference for DST being enabled or disabled" do
+      let(:jdbc_default_timezone) { 'America/Chicago' }
+
+      it 'blows up, honestly (legacy behaviour)' do
+        expect { plugin.run(queue) }.to raise_error(::Sequel::InvalidValue, /AmbiguousTime/)
+      end
+    end
+  end
+
   context "when an unreadable jdbc_driver_path entry is present" do
     let(:driver_jar_path) do
       jar_file = $CLASSPATH.find { |name| name.index(Jdbc::Derby.driver_jar) }
--- a/spec/inputs/jdbc_spec.rb
+++ b/spec/inputs/jdbc_spec.rb
@@ -1468,7 +1468,7 @@ describe LogStash::Inputs::Jdbc do
     end
 
     context "when initialized with a preference for DST being enabled" do
-      let(:jdbc_default_timezone) { 'America/Chicago[prefer-dst:true]' }
+      let(:jdbc_default_timezone) { 'America/Chicago[dst_enabled_on_overlap:true]' }
 
       it 'treats the timestamp column as if DST was enabled' do
         plugin.run(queue)
@@ -1477,7 +1477,7 @@ describe LogStash::Inputs::Jdbc do
       end
     end
     context "when initialized with a preference for DST being disabled" do
-      let(:jdbc_default_timezone) { 'America/Chicago[prefer-dst:false]' }
+      let(:jdbc_default_timezone) { 'America/Chicago[dst_enabled_on_overlap:false]' }
 
       it 'treats the timestamp column as if DST was disabled' do
         plugin.run(queue)
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -5,4 +5,4 @@ set -ex
 
 export USER='logstash'
 
-bundle exec rspec spec && bundle exec rspec spec --tag integration
+bundle exec rspec spec --format documentation && bundle exec rspec spec --format documentation --tag integration
--- a/lib/logstash/plugin_mixins/jdbc/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc/jdbc.rb
@@ -4,6 +4,7 @@ require "logstash/config/mixin"
 require "time"
 require "date"
 require_relative "value_tracking"
+require_relative "timezone_proxy"
 require_relative "checked_count_logger"
 require_relative "statement_handler"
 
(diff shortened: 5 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 5.2.0
+  - Added support for disambiguating timestamps in daylight-savings overlap periods
```
karenzone, line 2:
```suggestion
  - Added support for disambiguating timestamps in daylight saving time (DST) overlap periods [#92](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/92)
```

karenzone: The changelog entry gets picked up for release notes, and it's helpful to have a link to the PR.

karenzone (commented):
Docs review:  Nice explanations here! I offered some minor wording and formatting suggestions to make your content pop. 

Without formatting changes: 
Screen Shot 2021-12-01 at 3 47 34 PM

With formatting changes: 
Screen Shot 2021-12-01 at 4 42 59 PM

docs/input-jdbc.asciidoc
```diff
@@ -282,9 +282,13 @@ JDBC connection string
 ===== `jdbc_default_timezone`
 
   * Value type is <<string,string>>
+  ** Value should be a canonical timezone or offset, such as `Europe/Paris` or `Etc/GMT+3`
+  ** Value _may_ include square-bracketed extensions, such as `America/Denver[dst_enabled_on_overlap:true]`
   * There is no default value for this setting.
 
-Timezone conversion.
+[id="plugins-{type}s-{plugin}-jdbc_timezone_conv"]
+Timezone conversion
```
karenzone, line 290:
```suggestion
====== Timezone conversion
```

docs/input-jdbc.asciidoc
```diff
@@ -295,7 +299,15 @@ in relative UTC time in ISO8601 format.
 
 Using this setting will manually assign a specified timezone offset, instead
 of using the timezone setting of the local machine.  You must use a canonical
-timezone, *America/Denver*, for example.
+timezone, `America/Denver`, for example.
+
+[id="plugins-{type}s-{plugin}-jdbc_ambiguous_timestamps"]
+Ambiguous timestamps
```
karenzone, line 305:
```suggestion
===== Ambiguous timestamps
```

docs/input-jdbc.asciidoc
```diff
@@ -282,9 +282,13 @@ JDBC connection string
...
   * Value type is <<string,string>>
+  ** Value should be a canonical timezone or offset, such as `Europe/Paris` or `Etc/GMT+3`
+  ** Value _may_ include square-bracketed extensions, such as `America/Denver[dst_enabled_on_overlap:true]`
   * There is no default value for this setting.
 
-Timezone conversion.
+[id="plugins-{type}s-{plugin}-jdbc_timezone_conv"]
+Timezone conversion
+
 Logstash (and Elasticsearch) expects that timestamps are expressed in UTC terms.
```
karenzone, line 292:
```suggestion
Logstash and Elasticsearch expect timestamps to be expressed in UTC terms.
```

karenzone: I know that this sentence predated this PR and I understand why it was written this way. Still seems like a good opportunity to make an update. I don't think anybody who is not using ES will be confused by this statement, and it gives us an opportunity to clear up some grammar ambiguity.

karenzone (commented):
Left suggestions inline. Did you choose not to format these items as subheadings, or is that an oversight? FWIW, they will perform better in SEO rankings if they are headings. You make the call.

Otherwise, docs **LGTM**.

spec/inputs/jdbc_spec.rb
```diff
@@ -1447,6 +1447,53 @@
...
+      db << "INSERT INTO types_table (num, string, started_at, custom_time, ranking) VALUES (1, 'A test', '1999-12-31', '2021-11-07 01:23:45', 95.67)"
+
+      plugin.register
+
+      # plugin.set_value_tracker(value_tracker)
+      # allow(value_tracker).to receive(:value).and_return("bar")
+    end
+
+    context "when initialized with a preference for DST being enabled" do
+      let(:jdbc_default_timezone) { 'America/Chicago[dst_enabled_on_overlap:true]' }
```
kares, line 1471: what I like here that we're keeping the dst handling information close to the timezone, what I do not like is that we extend the TZ string I know it's unlikely but what happens if users request a different behavior or smt else needs to be extended in the time-zone.

another thing to consider, since time is ambiguous twice a year I wonder if users would ever want to have different behavior (once use dst, second time fallback to non dst time) - a bit far fetched, why anyone would want that, but we should keep our options open.

ways forward seems to be:
1. have a separate configuration option e.g. `timezone_dst_enabled_on_overlap => true/false`
    or smt of a `ambiguous_timezone_local_time_handling => use_dst|use_non_dst|error`
2. since the flag relates to time zone maybe extend the option to accepts a `string_or_hash` e.g.
```ruby
input {
  jdbc {
    jdbc_connection_string => "..."
    jdbc_default_timezone => {
      name => 'America/Chicago',
      
      ambiguous_local_time_handling => dst
      # or
      dst_enabled_on_overlap => true
      
    }
    use_column_value => true
    # ...
  }
}
```

3. extend the `jdbc_default_timezone` with custom parsing `[]` (current PR status e.g.  `America/Chicago[dst_enabled_on_overlap:true]`)

kares (commented):
:1st_place_medal: so great to finally see ambiguous DST timestamps being handled ...

no concerns implementation wise assuming we are in agreement with the way to name and configure the feature.

left a comment to consider our options above.

spec/inputs/jdbc_spec.rb
yaauie: I attempted and ruled out the first option (separate config entirely) because it is just too separate and users aren't likely to find it until _after_ they have run into a DST-transition issue.

I have not yet made a serious attempt at the second (`string_or_hash`). On the surface I like it, but I believe it would benefit from a [validator hack](https://github.com/logstash-plugins/logstash-mixin-validator_support/blob/main/lib/logstash/plugin_mixins/validator_support.rb) to ensure we keep error messages from malformed instructions close to the user. I will make an attempt at this, and determine if it is cleared in practice.

This PR implements the third option and attempts to remain open to future extension:
 - Named timezones are effectively constrained to [slashes (`/`), ASCII letters (`a-z` and `A-Z`), dashes (`-`), dots (`.`), and underscores (`_`)](https://data.iana.org/time-zones/theory.html#naming); by defining extensions enclosed in square brackets, we are immune to upstream changes to the timezonedb
 - Our implementation only uses two hard-coded extensions in order to avoid writing a complex parser, but would easily accept multiple semicolon-separated extensions.
 - After _many_ attempts at wording, I decided to stick with ruby stdlib's naming `dst_enabled_on_overlap`, because this extension _only_ handles DST-related ambiguity (and not ambiguity caused by political-upheaval).

yaauie: Whether we do the second or third approach, adding a validator brings errors closer to the user and is easy enough:

~~~ ruby
    module JDBCTimezoneSpecValidator
      def validate_value(value, validator_name)
        return super(value, validator_name) unless validator_name == :jdbc_timezone_spec

        return true, value if value.kind_of?(::TZInfo::Timezone) # multiple coercion passes
        return true, nil if value.nil? || (value.kind_of?(String) && value.empty?)

        [true, TimezoneProxy.load(value)] rescue [false, $!.message]
      end
    end
~~~

AS to the implementation of accepting either `jdbc_default_timezone => "America/Los_Angeles"` or `jdbc_default_timezone => { name => "America/Los_Angeles" dst_enabled_on_overlap => true }`, that is also straight-forward enough:

~~~ ruby
    def self.load(spec)
      return spec if spec.kind_of?(::TZInfo::Timezone)

      if spec.kind_of?(Hash)
        extensions = spec.dup
        name = extensions.delete("name") { fail(ArgumentError, "timezone must include 'name' key") }
      else
        name = spec
        extensions = {}
      end

      timezone = ::TZInfo::Timezone.get(name)

      if extensions && extensions.include?("dst_enabled_on_overlap")
        dst_enabled_on_overlap = extensions.delete("dst_enabled_on_overlap")
        case dst_enabled_on_overlap.to_s.downcase
        when 'true'  then timezone = timezone.dup.extend(PeriodForLocalWithDSTPreference::ON)
        when 'false' then timezone = timezone.dup.extend(PeriodForLocalWithDSTPreference::OFF)
        else fail(ArgumentError, "Invalid timezone extension `dst_enabled_on_overlap:#{dst_enabled_on_overlap}`")
        end
      end

      fail(ArgumentError, "Unexpected timezone extension: #{extensions}") unless extensions.empty?

      timezone
    end
~~~


Interface-wise, I think the square-bracket approach is a bit more straight-forward because the pipeline syntax for maps isn't always clear (whitespace separation, not commas):

~~~
input {
  jdbc {
    jdbc_connection_string => "..."
    jdbc_default_timezone => {
      name => "America/Los_Angeles"
      dst_enabled_on_overlap => true
    }
    use_column_value => true
    # ...
  }
}
~~~

VS

~~~
input {
  jdbc {
    jdbc_connection_string => "..."
    jdbc_default_timezone => "America/Los_Angeles[dst_enabled_on_overlap:true]"
    use_column_value => true
    # ...
  }
}
~~~

kares: thanks for looking into this. 
fine by shipping the current approach given that the risk of additional time-zone extensions is low (which would blow off the string e.g. `America/Los_Angeles[dst_enabled_on_overlap:true][another_extension_on_dst_switch:+1h]`) I have no further concerns.

spec/inputs/jdbc_spec.rb
```diff
@@ -1447,6 +1447,53 @@
...
+        expect(event.get("custom_time")).to be_a_logstash_timestamp_equivalent_to("2021-11-07T07:23:45Z")
+      end
+    end
+    context "when initialized without a preference for DST being enabled or disabled" do
+      before(:each) { allow(plugin.logger).to receive(:warn) }
+      let(:jdbc_default_timezone) { 'America/Chicago' }
+
+      it 'the error results in helpful log warning' do
+        # logger = plugin.instance_variable_get(:@logger)
+        # puts("LOGGER: METHOD(#{plugin.logger.inspect}) IVAR(#{logger.inspect})")
```
kares, line 1488: nit: let's clean up the commented bits
```suggestion
```

spec/inputs/jdbc_spec.rb
```diff
@@ -1447,6 +1447,53 @@
...
+        plugin.run(queue)
+        # expect(plugin.logger).to have_received(:warn) { |*actual_args| puts "WARN>>#{actual_args.inspect}" }
+        expect(plugin.logger).to have_received(:warn).with(a_string_including("Exception when executing JDBC query"), a_hash_including(:exception => a_string_including("2021-11-07 01:23:45 is an ambiguous local time")))
```
kares, line 1491: :heart_eyes: the assert, would the commented bits be still useful later on?

kares (approved):
:nail_care: how the implementation wraps the `TZInfo::Timezone` object,
this allows multiple jdbc plugins to function in isolation as expected (opposed to setting a global `Sequel.tzinfo_disambiguator`)

spec/inputs/jdbc_spec.rb
```diff
@@ -1447,6 +1447,53 @@
...
+    end
+    context "when initialized without a preference for DST being enabled or disabled" do
+      before(:each) { allow(plugin.logger).to receive(:warn) }
+      let(:jdbc_default_timezone) { 'America/Chicago' }
+
+      it 'the error results in helpful log warning' do
+        # logger = plugin.instance_variable_get(:@logger)
+        # puts("LOGGER: METHOD(#{plugin.logger.inspect}) IVAR(#{logger.inspect})")
+        plugin.run(queue)
+        # expect(plugin.logger).to have_received(:warn) { |*actual_args| puts "WARN>>#{actual_args.inspect}" }
```
yaauie, line 1490:
```suggestion
```

spec/inputs/jdbc_spec.rb
```diff
@@ -1447,6 +1447,53 @@
...
+        expect(plugin.logger).to have_received(:warn).with(a_string_including("Exception when executing JDBC query"), a_hash_including(:exception => a_string_including("2021-11-07 01:23:45 is an ambiguous local time")))
+        # expect { plugin.run(queue) }.to_not raise_error(::Sequel::InvalidValue, /AmbiguousTime/)
```
yaauie, line 1492:
```suggestion
```

yaauie:
Changes since last approval:
 - merging `upstream/main` [17c4123](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/92/commits/17c412301f3fcaf6095bc7459b15f450da4bb385), conflicts resolved by:
   - composing changelog
   - bumping version appropriately
   - removing code that was previously removed from `main` in https://github.com/logstash-plugins/logstash-integration-jdbc/pull/109
 - minor refinement of specs to match actual user input
 - bump gradle to support Java 17 (and therefore Logstash 8.4+)
 - fix warnings about using `BigDecimal` in specs

roaksoax (approved):
LGTM!

PR #94, opened 2021-12-09
PR #90 support

andsel:
This PR is support of #90 which is created on the @dingx1 `master` branch and not on a dedicated PR branch


andsel:
Superseeded by #95

PR #95, opened 2021-12-10
Add another flexible way for JDBC paging (manual mode)

dingx1:
## Release notes

Added `jdbc_paging_mode` option to choose if use `explicit` pagination in statements and avoid the initial count query or use `auto` to delegate to the underlying library.

## What does this PR do?

This commit introduces a new configuration option to avoid the initial count select statement executed by Sequel in case of paginated queries. 
The paginated queries could be done now with initial count or not, in case `jdbc_paging_mode` is values `explicit` the plugin executes paged queries till it reach a page with less rows than the expected, instead of relying on the total row count.
In this case the SQL statement has to explicitly use the pagination keywords (like `LIMIT` and `OFFSET`) receiving as `:offset` and `:size` as implicit parameters.
When the paging mode `jdbc_paging_mode` is set to `auto` then the pagination happens automatically without the intervention of the user to create paginated query.

In some circumstances, like stored procedure call with pagination bounds, the initial count query is simply not meaningful.
In cases where the pages to retrieve are few the initial count query could be an useful overhead.
In some complex nested queries the count could generate extra work that's not usefull

## Checklist

- [x] I have made corresponding changes to the documentation

## Author's Checklist

- [x] check it with a local Logstash test.

- spin up a local DB
- configure a JDBC input query with pagination and with the new `jdbc_paging_avoid_count` and check it retrieves all the expected rows. In logs should be present a line for each paged query

```
input {
  jdbc {
    jdbc_driver_library => "/path/mysql-connector-java-8.0.26.jar"
    jdbc_driver_class => "Java::com.mysql.cj.jdbc.Driver"
    jdbc_connection_string => "jdbc:mysql://localhost:3306/test_logstash"

    jdbc_user => "user"
    jdbc_password => "s3cret"

    jdbc_default_timezone => "UTC"

    schedule => "* * * * *"

    use_column_value => true
    tracking_column => "log_id"
    tracking_column_type => "numeric"
    last_run_metadata_path => ".last_run"
    
    statement => "SELECT * FROM data_log WHERE log_id > :sql_last_value LIMIT :size OFFSET :offset"
    jdbc_paging_enabled => true
    jdbc_page_size => 7
    jdbc_paging_avoid_count => true
  }
}

output {
  stdout {
    codec => "rubydebug"
  }
}
```

## Related issues

- Superseeds #90

### Simple query with small result sets
In a simple query like:
```sql
SELECT * from test_table
```
the automatic pagination would generate the count query plus the paginated queries.
With small resultsets it issues 2 queries instead of one query.

### A stored procedure that requires pagination information. Suppose you have a stored like
 ```sql
CALL fetch_my_data(:sql_last_value, <offset>, <size>)
```
the automatic pagination would generate an invalid SQL statement:
```sql
SELECT count(*) AS count FROM ( CALL fetch_my_data(:sql_last_value) ) AS t1 LIMIT 1
SELECT * FROM ( CALL fetch_my_data(:sql_last_value) ) AS t1 LIMIT 10 OFFSET 0
.
.
```
which is not what's expected.

### Nested queries with pagination on the inner query
Suppose the query you want to paginate is the inner on, like in:
```sql
SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table LIMIT <offset>, <size>)
```

Without this feature, the automatic pagination would generate the following SQL:
```sql
SELECT count(*) AS count FROM ( 
  SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table) 
) AS t1 LIMIT 1;

SELECT * FROM ( 
  SELECT * FROM test_table2 WHERE id IN (SELECT * FROM test_table) 
) AS t1 LIMIT 10 OFFSET 0;
.
.
```
which is the pagination applied to the outer query instead of the inner one.

### DIFF
--- a/docs/input-jdbc.asciidoc
+++ b/docs/input-jdbc.asciidoc
@@ -385,12 +385,12 @@ Whether to use count query during the JDBC paging
 If false, a count query followed by multiple paged queries (with `LIMIT` statement, etc.) are used under the hood.
 
 If true, multiple paged queries (without a count query ahead) will be performed until no more rows are retrieved.
-The `offset` and `size` parameters can be used in the statement (`size` equal to `jdbc_page_size`, and `offset` incremented by `size`).
+The `offset` and `size` parameters must be used in your statement (`size` equal to `jdbc_page_size`, and `offset` incremented by `size`).
 Example:
 
 [source, ruby]
 ------------------------------------------------------
-"SELECT id, mycolumn1, mycolumn2 FROM my_table WHERE id > :sql_last_value LIMIT :offset, :size"
+"SELECT id, mycolumn1, mycolumn2 FROM my_table WHERE id > :sql_last_value LIMIT :size OFFSET :offset"
 ------------------------------------------------------
 
 [source, ruby]
--- a/docs/input-jdbc.asciidoc
+++ b/docs/input-jdbc.asciidoc
@@ -153,7 +153,7 @@ NOTE: Not all JDBC accessible technologies will support prepared statements.
 With the introduction of Prepared Statement support comes a different code execution path and some new settings. Most of the existing settings are still useful but there are several new settings for Prepared Statements to read up on.
 Use the boolean setting `use_prepared_statements` to enable this execution mode. Use the `prepared_statement_name` setting to specify a name for the Prepared Statement, this identifies the prepared statement locally and remotely and it should be unique in your config and on the database. Use the `prepared_statement_bind_values` array setting to specify the bind values, use the exact string `:sql_last_value` (multiple times if necessary) for the predefined parameter mentioned before. The `statement` (or `statement_path`) setting still holds the SQL statement but to use bind variables you must use the `?` character as a placeholder in the exact order found in the `prepared_statement_bind_values` array.
 
-NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries (when `jdbc_page_size` is greater than `0`) under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
+NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries (when `jdbc_paging_avoid_count` is false) under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
 
 Example:
 [source,ruby]
@@ -193,6 +193,7 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 | <<plugins-{type}s-{plugin}-jdbc_fetch_size>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-jdbc_page_size>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-jdbc_paging_enabled>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-jdbc_paging_avoid_count>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-jdbc_password>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-jdbc_password_filepath>> |a valid filesystem path|No
 | <<plugins-{type}s-{plugin}-jdbc_pool_timeout>> |<<number,number>>|No
@@ -359,22 +360,6 @@ JDBC fetch size. if not provided, respective driver's default will be used
 
 JDBC page size
 
-If greater than `0`, a count query and multiple paged queries (with `LIMIT` statement, etc.) are used under the hood.
-
-Otherwise, if less than `0`, multiple queries (without count query) will be performed until no rows are returned.
-The `:offset` and `:size` parameters can be used in the statement (`:size` equal to `-jdbc_page_size`, and `:offset` incremented by `:size`).
-Example:
-
-[source, ruby]
------------------------------------------------
-"SELECT id, mycolumn1, mycolumn2 FROM my_table WHERE id > :sql_last_value LIMIT :offset, :size"
------------------------------------------------
-
-[source, ruby]
------------------------------------------------
-"CALL fetch_my_data(:sql_last_value, :offset, :size)"
------------------------------------------------
-
 [id="plugins-{type}s-{plugin}-jdbc_paging_enabled"]
 ===== `jdbc_paging_enabled`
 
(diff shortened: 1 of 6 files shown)


andsel (approved):
LGTM

docs/input-jdbc.asciidoc
```diff
@@ -129,6 +129,9 @@ Here is the list:
 |sql_last_value | The value used to calculate which rows to query. Before any query is run,
 this is set to Thursday, 1 January 1970, or 0 if `use_column_value` is true and
 `tracking_column` is set. It is updated accordingly after subsequent queries are run.
+|offset, size| Values used in manual paging mode. Used in queries to explicitly implement
+the paging. Usable only if <<plugins-{type}s-{plugin}-jdbc_paging_enabled>> and
+<<plugins-{type}s-{plugin}-jdbc_paging_manual_mode>>  are enabled.
```
andsel, line 134: @karenzone I would ask for your eyes on this, please :-)

lib/logstash/plugin_mixins/jdbc/jdbc.rb
```diff
@@ -55,6 +55,9 @@ def setup_jdbc_config
       # Be aware that ordering is not guaranteed between queries.
       config :jdbc_paging_enabled, :validate => :boolean, :default => false
 
+      # Whether to use manual mode during the JDBC paging
+      config :jdbc_paging_manual_mode, :validate => :boolean, :default => false
```
kares, line 59: sorry for the late review, did you guys consider having better naming for the feature? e.g.

`jdbc_paging_enabled => true jdbc_paging_mode => offset` any chance this reads better (than `"manual"`)?
(`jdbc_paging_mode => count` would de the default value)

should also keep things open for changing the paging strategy again (offset paging is a bit of an anti pattern although for mostly static data and LS' needs of at least once delivery it's sufficient), wdyt?

andsel: @kares I like your suggestion,  instead of using a boolean flag defined a string value for the strategy is more readable.
However this feature doesn't cover only static data but also use cases where the `Sequel` automatic paging doesn't apply correctly, mostly:
- procedure calls with paging limits
- paging in inner queries in case of nested SQL queries

kares: I see, maybe we need better naming there - haven't thought about it that much.
however the manual just read a bit weird - had no idea what that could mean SQL wise.

what I did is check Sequel's [`each_page`](https://github.com/evanfarrar/opensprints/blob/master/lib/sequel/lib/sequel/extensions/pagination.rb#L22-L24) and it's using a `COUNT(...)` initial query (the details on how the statement is transformed into a count is usually adapter specific), haven't checked any adapter specific overrides for `each_page`.

andsel: @kares to me the naming you are proposing is good, better than the boolean flag. So I would use it.

docs/input-jdbc.asciidoc
```diff
@@ -129,6 +129,9 @@ Here is the list:
 |sql_last_value | The value used to calculate which rows to query. Before any query is run,
 this is set to Thursday, 1 January 1970, or 0 if `use_column_value` is true and
 `tracking_column` is set. It is updated accordingly after subsequent queries are run.
+|offset, size| Values used in manual paging mode. Used in queries to explicitly implement
```
karenzone, line 132: Weird... github won't let me re-select those 3 rows so that I can make a suggestion.  So what about this: 

|offset, size| Values used with manual paging mode to explicitly implement
the paging. 
Supported only if <<plugins-{type}s-{plugin}-jdbc_paging_enabled>> and
<<plugins-{type}s-{plugin}-jdbc_paging_manual_mode>> are enabled.

I think it's OK not to repeat that this applies to queries as that is clearly stated in the lead-in sentence before the bullets.  That leaves us with a tighter description.   WDYT?

docs/input-jdbc.asciidoc
karenzone: ^^ Thanks for a chance to take a look. See comment at line 132. :-)

docs/input-jdbc.asciidoc
andsel: Reads far better, I'll use it!

lib/logstash/plugin_mixins/jdbc/jdbc.rb
dingx1: @kares @andsel @karenzone Thanks for your time. I also think this suggestion is good and will improve according to it.
I have a little question about the name.
Actually, this feature involves two options:
1. Not to add paging condition automatically, but explicitly customized by user.
1. Not to use count query first

The first one is the main purpose of this feature.
So, is it easy for users to understand the two values of 'count' and 'offset'?
If everyone don't have other suggested names, I will implement it according to current suggestion.

andsel: Good point, we could use 
- `automatic` for `dbc_paging_mode => count` and 
- `explicit` for `jdbc_paging_mode => offset`

@kares is it a better naming in your opinion?

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -27,14 +27,30 @@ def post_init(plugin)
...
     # Performs the query, respecting our pagination settings, yielding once per row of data
     # @param db [Sequel::Database]
-    # @param sql_last_value [Integet|DateTime|Time]
+    # @param sql_last_value [Integer|DateTime|Time]
+    # @param jdbc_paging_enabled [Boolean]
+    # @param jdbc_paging_manual_mode [Boolean]
+    # @param jdbc_page_size [Integer]
     # @yieldparam row [Hash{Symbol=>Object}]
-    def perform_query(db, sql_last_value, jdbc_paging_enabled, jdbc_page_size)
+    def perform_query(db, sql_last_value, jdbc_paging_enabled, jdbc_paging_manual_mode, jdbc_page_size)
```
kares, line 35: would be nice if there was a new class as it's a different algorithm e.g. `ExplicitStatementHandler` instead of hacking the algorithm into the existing one.

lib/logstash/plugin_mixins/jdbc/jdbc.rb
kares: those sound better instead of automatic maybe `jdbc_paging_mode => auto` to choose the best available strategy based on the driver (Sequel does this). `jdbc_paging_mode => explicit` or `manual` both sounds okay to me.

kares (changes requested):

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
andsel: nice idea, I come up with some commits to switch this

andsel:
@kares I've integrated your great latest suggest, I would ask another round of your eyes on this, please.

kares (commented):
LGTM, well done!

andsel:
Many thanks @dingx1 for all of this :1st_place_medal:  it was published with https://rubygems.org/gems/logstash-integration-jdbc/versions/5.2.0

dingx1:
Wow, it's finally done. Great! ^_^
Thank you @andsel and everyone for your help！

Issue #83, opened 2021-09-28
Intermittent error in loading multiple drivers when SyBase driver is used


andsel:
**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`) `7.14.1` but it happens also in `master`
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker) tar.gz / sources
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes) command line
4. How was the Logstash Plugin installed, already shipped

**JVM** (e.g. `java -version`): 

If the affected version of Logstash is 7.9 (or earlier), or if it is NOT using the bundled JDK or using the 'no-jdk' version in 7.10 (or higher), please provide the following information:

1. JVM version (`java -version`)
```
openjdk version "11.0.11" 2021-04-20
OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK-11.0.11+9 (build 11.0.11+9, mixed mode)
```
2. JVM installation source (e.g. from the Operating System's package manager, from source, etc). from source
3. Value of the `JAVA_HOME` environment variable if set. `/home/andrea/.sdkman/candidates/java/current`

**OS version** (`uname -a` if on a Unix-like system):
`Linux kalispera 5.4.0-86-generic #97-Ubuntu SMP Fri Sep 17 19:19:40 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

**Description of the problem including expected versus actual behavior**:
When multiple pipelines with JDBC input configured to pull data from different DBs are present and one of them is `SyBase, jConn4.jar` then during startup it happens then randomly some pipelines have problems in loading the driver, and iterate in loading again and again but without success:
```
[2021-09-28T10:05:31,030][INFO ][logstash.javapipeline    ][oracle_1] Pipeline Java execution initialization time {"seconds"=>0.38}
[2021-09-28T10:05:31,030][INFO ][logstash.javapipeline    ][sybase_1] Pipeline Java execution initialization time {"seconds"=>0.38}
[2021-09-28T10:05:31,030][INFO ][logstash.javapipeline    ][mysql_1] Pipeline Java execution initialization time {"seconds"=>0.29}
[...]

PR #96, opened 2021-12-13
Refactor: to explicit Java (driver) class name loading

kares:
This is a second attempt on fixing https://github.com/logstash-plugins/logstash-integration-jdbc/issues/83 as it seems to still manifest even with locking.

#### Few more observations on what's going on (with #83):
- [`Sequel::JDBC.load_driver`](https://github.com/jeremyevans/sequel/blob/5.51.0/lib/sequel/adapters/jdbc.rb#L55) ends up doing an `eval` on the input (e.g. `Java::oracle.jdbc.driver.OracleDriver`)
- the `eval` sometimes fails with a `NameError` (having multiple drivers and executing threads is a factor)
- as hinted previously the problem seems to be specific to `'com.sybase.jdbc4.jdbc.SybDriver'` driver (+ having another driver such as `'oracle.jdbc.driver.OracleDriver'` around as well)
- JRuby might have an underlying bug with `eval` and setting up the Java package structure for the drivers
- at [one point an ugly error came out](https://gist.github.com/kares/850767d4be4d5323b068af9585e45d4a#file-load_driver_xfail-txt-L18-L35) - which indicates an underlying Java driver loading issue (although this might have been caused by the concurrent loading of different drivers - registering with `DriverManager`)

- the issue **only happens when drivers are on the LS class-path** (setting `jdbc_driver_library =>` did not reproduce!)
- with the changes in this PR the issue **did no longer reproduce even if drivers were on the LS class-path**

A bit simplified reproducer (does not need to run Logstash): https://gist.github.com/kares/850767d4be4d5323b068af9585e45d4a

*p.s. Previous changes (from https://github.com/logstash-plugins/logstash-integration-jdbc/pull/84) could be reverted but I am keeping them in (they do not hurt - the lock just slows down multiple jdbc plugins starting), for now.*

---

More investigation might be needed for an actual underlying cause (as well as the potential JRuby bug) but the changes presented here make sense regardless - skipping `eval` and doing a simply `Class.forName` is very similar to what an actual Java JDBC application would attempt to do (and should hopefully make debugging the problem easier).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.1.9
+  - Refactor: to explicit Java (driver) class name loading [#96](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/96),
+    the change is expected to provide a more robust fix for the driver loading issue [#83](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/83).
+
 ## 5.1.8
   - Fix the blocking pipeline reload and shutdown when connectivity issues happen [#85](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/85)
 
--- a/lib/logstash/plugin_mixins/jdbc/common.rb
+++ b/lib/logstash/plugin_mixins/jdbc/common.rb
@@ -4,6 +4,8 @@ module LogStash module PluginMixins module Jdbc
 
     private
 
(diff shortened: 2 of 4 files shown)


andsel (approved):
LGTM
The red CI is due to https://github.com/logstash-plugins/.ci/issues/31

Nice idea the one don't use the `Sequel` `load_driver` method and go straight to the Java way of loading.

PR #97, opened 2021-12-15
Fix: compatibility with all (>= 3.0) rufus-scheduler versions

kares:
The work here is meant to get the plugin working with up-to-date **rufus-scheduler** versions.

ATM, LS 7.x is stuck with gem version **3.0.9** (due plugins pinning `~> 3.0.9`), see https://github.com/elastic/logstash/issues/12931

The previous *rufus-scheduler* `< 3.5` lock (got tightened in version 3.2.2 to `~> 3.0.9`), was due Rufus moving the `Rufus::Scheduler::CronLine` class into a [standalone gem](https://github.com/floraison/fugit) (`Fugit::Cron`) with a slightly changed API (removed `#days` method). That feature was only used for logging and is now replaced by simply logging the calculated frequency of the cron expression.

Also, part of the PR our custom `Scheduler` sub-class was reviewed for re-use with a new API draft to use across plugins (as we plan to extract the `Scheduler` mixin bits into its own gem), the idea being using a factory method to provide everything in one step e.g.
```ruby
      @scheduler = ...::Scheduler.
          start_cron_scheduler(schedule, thread_name: "[#{id}]<...") { do_periodic_work(queue) }
      @scheduler.join
```
Most plugins will only need ^^^ this, the JDBC plugin still rely on some `Rufus::Scheduler` api (not worth to push the scheduler externalization that much further atm).

~~*NOTE: we should consider not using any 3.1 - 3.5 version which had stopped using **tzinfo** gem for a self invented `Rufus::ZoTime` class, tzinfo is (transitively) used again in version >= 3.6 through an added gem.*~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,7 @@
 import:
-- logstash-plugins/.ci:travis/travis.yml@1.x
+  - logstash-plugins/.ci:travis/travis.yml@1.x
+
+env:
+  jobs: # test with old scheduler version (3.0 was locked in LS 7.x)
+    - ELASTIC_STACK_VERSION=7.x RUFUS_SCHEDULER_VERSION=3.0.9
+    - ELASTIC_STACK_VERSION=6.x RUFUS_SCHEDULER_VERSION=3.0.9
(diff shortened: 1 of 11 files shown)


andsel (approved):
LGTM!

A nice work in moving the Scheduler code in a common mixin

PR #98, opened 2021-12-16
Fix: regression due returning the Java driver class

kares:
as not all drivers register with `java.sql.DriverManager` upon class initialization we need to account for `Sequel`'s way of fallback to instantiate a driver class using: `klass.new.connect(...)` (opposed to `klass.newInstance` which would have worked in this case).

NOTE: our tests did not catch this regression from https://github.com/logstash-plugins/logstash-integration-jdbc/pull/96 ...

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/common.rb
+++ b/lib/logstash/plugin_mixins/jdbc/common.rb
@@ -1,3 +1,4 @@
+require 'jruby'
 
 module LogStash module PluginMixins module Jdbc
   module Common
@@ -75,10 +76,13 @@ module LogStash module PluginMixins module Jdbc
 
     def load_jdbc_driver_class
       # sub a potential: 'Java::org::my.Driver' to 'org.my.Driver'
-      class_name = @jdbc_driver_class.gsub('::', '.').sub(/^Java\./, '')
+      klass = @jdbc_driver_class.gsub('::', '.').sub(/^Java\./, '')
       # NOTE: JRuby's Java::JavaClass.for_name which considers the custom class-loader(s)
       # in 9.3 the API changed and thus to avoid surprises we go down to the Java API :
-      JRuby.runtime.getJavaSupport.loadJavaClass(class_name) # throws ClassNotFoundException
+      klass = JRuby.runtime.getJavaSupport.loadJavaClass(klass) # throws ClassNotFoundException
+      # unfortunately we can not simply return the wrapped java.lang.Class instance as
+      # Sequel assumes to be able to do a `driver_class.new` which only works on the proxy,
+      org.jruby.javasupport.Java.getProxyClass(JRuby.runtime, klass)
     end
 
   end


CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-## 5.1.9
+## 5.1.10
+  - Refactor: to explicit Java (driver) class name loading [#96](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/96),
```
kares, line 2: same changelog as in 5.1.9 as the gem version has been yanked ...

lib/logstash/plugin_mixins/jdbc/common.rb
```diff
@@ -75,10 +76,13 @@ def jdbc_driver_library_set?
 
     def load_jdbc_driver_class
       # sub a potential: 'Java::org::my.Driver' to 'org.my.Driver'
-      class_name = @jdbc_driver_class.gsub('::', '.').sub(/^Java\./, '')
+      klass = @jdbc_driver_class.gsub('::', '.').sub(/^Java\./, '')
       # NOTE: JRuby's Java::JavaClass.for_name which considers the custom class-loader(s)
       # in 9.3 the API changed and thus to avoid surprises we go down to the Java API :
-      JRuby.runtime.getJavaSupport.loadJavaClass(class_name) # throws ClassNotFoundException
+      klass = JRuby.runtime.getJavaSupport.loadJavaClass(klass) # throws ClassNotFoundException
```
andsel, line 82: In PR #96 we switched from  `Sequel::JDBC.load_driver(` https://github.com/logstash-plugins/logstash-integration-jdbc/pull/96/files#diff-b5cd0f75e4e795a8e01458b0edcc88b3027f141e3095524ddf663f0fd4ec77fdL33
to `JRuby.runtime.getJavaSupport.loadJavaClass(class_name)` and given that the problem is related to `SyBase` and `Oracle` drivers being present in the same classpath as cited [here](https://userapps.support.sap.com/sap/support/knowledge/en/2827462), why we simply don't switch back to the Sequel load_driver ?

andsel (commented):
I've a question about why we don't rollbacked to Sequel way to load drivers.

lib/logstash/plugin_mixins/jdbc/common.rb
kares: considered switching back as well (and I do not mind doing so if it's preferred), here's my reasoning to not switch back:
- the native Java exception is mapped to a `NameError` (by JRuby) but that than is [mapped](https://github.com/jeremyevans/sequel/blob/5.51.0/lib/sequel/adapters/jdbc.rb#L67-L68) to a Sequel error - we loose the error message at minimum
- on java.lang.Error we likely won't to fail miserably (which the Sequel way might not - or at least it's not clear)
- the code relies on `eval` which is fine but unnecessary - might mask some JRuby bugs

andsel: Thanks for claryfing

andsel (approved):
LGTM

PR #100, opened 2021-12-21
Chore: a note on the driver loading fix

kares:
NOTE: on what the #98 and #96 actually does to address the infinite pipeline (driver) loop issue.


andsel (approved):
LGTM

PR #102, opened 2022-01-17
Feat: name scheduler threads + redirect error logging

kares:
The gist of the work here is to avoid `STDERR.puts` from `Rufus::Scheduler#on_error` as well as
to set a `Thread#name` (which than sets Java thread name) for threads started by rufus-scheduler.

Threads started by a particular plugin will use a common prefix - making them identifiable from Java thread dumps:
- `[#{id}]<jdbc__scheduler` is used for the scheduler thread
- `[#{id}]<jdbc__scheduler_worker-00` will be the name of the first started worker thread

*NOTE: extracted from https://github.com/logstash-plugins/logstash-integration-jdbc/pull/97*

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/scheduler.rb
+++ b/lib/logstash/plugin_mixins/jdbc/scheduler.rb
@@ -1,0 +1,111 @@
+require 'rufus/scheduler'
+
+require 'logstash/util/loggable'
+
+module LogStash module PluginMixins module Jdbc
+  class Scheduler < Rufus::Scheduler
+
+    include LogStash::Util::Loggable
+
+    # Rufus::Scheduler >= 3.4 moved the Time impl into a gem EoTime = ::EtOrbi::EoTime`
+    # Rufus::Scheduler 3.1 - 3.3 using it's own Time impl `Rufus::Scheduler::ZoTime`
+    TimeImpl = defined?(Rufus::Scheduler::EoTime) ? Rufus::Scheduler::EoTime :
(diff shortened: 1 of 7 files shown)


lib/logstash/plugin_mixins/jdbc/scheduler.rb
```diff
@@ -0,0 +1,111 @@
...
+
+module LogStash module PluginMixins module Jdbc
+  class Scheduler < Rufus::Scheduler
+
+    include LogStash::Util::Loggable
+
+    # Rufus::Scheduler >= 3.4 moved the Time impl into a gem EoTime = ::EtOrbi::EoTime`
+    # Rufus::Scheduler 3.1 - 3.3 using it's own Time impl `Rufus::Scheduler::ZoTime`
+    TimeImpl = defined?(Rufus::Scheduler::EoTime) ? Rufus::Scheduler::EoTime :
+                   (defined?(Rufus::Scheduler::ZoTime) ? Rufus::Scheduler::ZoTime : ::Time)
```
kares, line 13: for now LS is locked to rufus-scheduler `3.0.9` but this is compatible with any version `>= 3.0.9`
(there's an upgrade pending at: https://github.com/logstash-plugins/logstash-integration-jdbc/pull/97)

andsel (approved):
LGTM!

Tried locally with LS `7.16.1`, building the gem and installing, and on launched LS 
```
jstack <LS_pid> | grep -i "jdbc__"
```

Issue #88, opened 2021-10-29
crond job successful execute a few times, but stuck without any error log later


szmengran:
config like this:
```
input {
  jdbc {
    jdbc_connection_string => "jdbc:mysql://xxx/xxx"
    jdbc_user => "xxx"
    jdbc_password => "xxx"
    jdbc_validate_connection => true
    jdbc_driver_library => "./logstash-core/lib/jars/mysql-connector-java.jar"
    jdbc_driver_class => "com.mysql.jdbc.Driver"
    schedule => "*/1 * * * * Asia/Shanghai"
    statement => "select * from memberbaseinfo a where lastlogintime > :sql_last_value"
    clean_run => false
    last_run_metadata_path => "./pipeline/.logstash_jdbc_last_run"
    plugin_timezone => "local"
    jdbc_default_timezone => "Asia/Shanghai"
  }
}
output {
  elasticsearch {
    hosts => ["xxx:9200"]
    action=>"update"
    index => "tags"
    doc_as_upsert => true
    document_id => "%{memberid}"
  }
  stdout {
    codec => json_lines
  }
}
```
The results are as follows：
```
[2021-10-28T18:51:08,490][INFO ][logstash.inputs.jdbc     ] (0.807486s) select a.memberid, a.nickname, a.sex, a.birthday, a.marriage, a.workcity, a.salary, a.education, ((1<<9 & health)>>9) ^ if(membertype>5, 2, 0) type, lastlogintime from memberbaseinfo a where lastlogintime > '2021-10-28 18:50:01'
{"@version":"1","birthday":"1993-12-24T14:57:02.000Z","nickname":"xx","workcity":10126099,"type":0,"@timestamp":"2021-10-28T10:51:08.495Z","sex":0,"education":5,"salary":5,"marriage":1,"lastlogintime":"2022-02-02T03:59:33.000Z","memberid":1828694900}
[2021-10-28T18:52:02,739][INFO ][logstash.inputs.jdbc     ] (0.363404s) select a.memberid, a.nickname, a.sex, a.birthday, a.marriage, a.workcity, a.salary, a.education, ((1<<9 & health)>>9) ^ if(membertype>5, 2, 0) type, lastlogintime from memberbaseinfo a where lastlogintime > '2021-10-28 18:51:07'
[...]

PR #103, opened 2022-02-10
Performance: avoid contention on scheduler execution

kares:
What we're after here is to avoid `Thread.list` blocking operation, which in JRuby means walking a synchronized Map.

A lot of Ruby threads in the system trying to `Thread.list` will lead to thread contending (blocked) for execution ...

`Rufus::Scheduler` uses `Thread.list` when figuring out which are the worker/scheduler threads it started.
This is working but a bit unfortunate and might be the cause for issues such as https://github.com/logstash-plugins/logstash-integration-jdbc/issues/88 (if there are multiple scheduler instances started from plugins - e.g. multiple jdbc inputs or http_poller or elasticsearch inputs with a `schedule => ...`).

Implementation wise there's 3 changes that improve performance:

1. less [scheduler thread ticks](https://github.com/jmettraux/rufus-scheduler/blob/v3.0.9/lib/rufus/scheduler.rb#L564) (sleep for 1.0 instead of 0.3)
Screenshot from 2022-02-09 14-29-08

2. avoid [timeout_jobs](https://github.com/jmettraux/rufus-scheduler/blob/v3.0.9/lib/rufus/scheduler.rb#L562) doing a `Thread.list` if the timeout feature is not used (the plugin does not use timeouts)
Screenshot from 2022-02-09 14-45-33
NOTE: this accounts for most of the improvements ...

3. finally there's still a `list` operation whenever the [job is about to be executed](https://github.com/jmettraux/rufus-scheduler/blob/v3.0.9/lib/rufus/scheduler/jobs.rb#L323)
Screenshot from 2022-02-09 15-48-48
NOTE: the reason we re-def `worker_threads` and cache them - there's no more locking operations (the remaining Random contention is due to using `Kernel.rand` in a local testing scenario and does not happen in production).

The recording screenshots are from a local emulated run, which is a bit far fetched: 600 schedulers with a 10s interval cron schedule. Still do not have a good enough explanation for a complete scheduler stop, what I was able to accomplish locally is some schedulers skipping one, two, sometimes three four (10 second) runs.

---

For now we're not sure whether there is another bug with `Rufus::Scheduler` (3.0.9) that would cause #88.

The plan is to have the patch tested with potentially many JDBC inputs, if the input stopping issue persist we might need to investigate further. Upon succession the `< Scheduler` sub-class should be extracted into it's own mixin to be usable by all plugins (not just JDBC), ideally the rufus-scheduler would also be updated to latest version.
The overrides should be iterated upon to avoid using `Thread.list` in the library completely and submitting a patch upstream.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/scheduler.rb
+++ b/lib/logstash/plugin_mixins/jdbc/scheduler.rb
@@ -27,7 +27,7 @@ module LogStash module PluginMixins module Jdbc
 
     # @overload
     def work_threads(query = :all)
-      if query.nil? # our special case from JobDecorator#start_work_thread
+      if query == :__all_no_cache__ # special case from JobDecorator#start_work_thread
         @_work_threads = nil # when a new worker thread is being added reset
         return super(:all)
       end
@@ -108,14 +108,14 @@ module LogStash module PluginMixins module Jdbc
     module JobDecorator
 
       def start_work_thread
-        prev_thread_count = @scheduler.work_threads(nil).size
+        prev_thread_count = @scheduler.work_threads.size
 
         ret = super() # does not return Thread instance in 3.0
 
-        work_threads = @scheduler.work_threads(nil)
+        work_threads = @scheduler.work_threads(:__all_no_cache__)
         while prev_thread_count == work_threads.size # very unlikely
           Thread.pass
-          work_threads = @scheduler.work_threads(nil)
+          work_threads = @scheduler.work_threads(:__all_no_cache__)
         end
 
         work_thread_name_prefix = @scheduler.work_thread_name_prefix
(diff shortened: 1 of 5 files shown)


lib/logstash/plugin_mixins/jdbc/scheduler.rb
```diff
@@ -12,6 +12,42 @@ class Scheduler < Rufus::Scheduler
...
+      # However, this isn't necessary we can easily detect whether there are any jobs
+      # that might need to timeout: only when `@opts[:timeout]` is set causes worker thread(s)
+      # to have a `Thread.current[:rufus_scheduler_timeout]` that is not nil
+      return unless @opts[:timeout]
+      super
+    end
+
+    # @overload
+    def work_threads(query = :all)
+      if query.nil? # our special case from JobDecorator#start_work_thread
```
andsel, line 30: Here we are using the `nil` value as a special value, we are assigning to it a meaning. `work_threads(nil)` is invoked only by the patched methods in `JobDecorator`, could we use something really specific and explicit for this?
Maybe adding a new label `:os_list_threads` or something similar instead of `nil` ?

andsel (changes requested):
Overall the ideas LGTM, have left only a (maybe) stylistic concern about the usage of `nil` as special case param.

andsel (approved):
LGTM !

PR #105, opened 2022-03-03
Removes double creation of FileHandler used in value tracking

andsel:
This commits removes redundant creation of FileHandler instance.


lib/logstash/plugin_mixins/jdbc/value_tracking.rb
```diff
@@ -6,9 +6,6 @@ class ValueTracking
 
     def self.build_last_value_tracker(plugin)
       handler = plugin.record_last_run ? FileHandler.new(plugin.last_run_metadata_path) : NullFileHandler.new(plugin.last_run_metadata_path)
-      if plugin.record_last_run
-        handler = FileHandler.new(plugin.last_run_metadata_path)
```
mashhurs, line 10: What is the current behavior if `FileHandler.new(plugin.last_run_metadata_path)` runs twice?

mashhurs (approved):

PR #106, opened 2022-03-04
Updates storage path for internal data to use subfolder in Logstash's `data.path`

andsel:
## Release notes

Change default path of 'last_run_metadata_path' to be rooted in the LS `data.path` folder and not in  $HOME. Also apply a path forward action to move existing metadata file to the new destination.

## What does this PR do?

Change default path of 'last_run_metadata_path' to be rooted in the LS data.path folder and not in  $HOME, in the input plugin.
Provide a path forward for existing metada file in `$HOME` to be moved in the new `data.path` destination.

Updates the path used by derby inside the jdbc_static filter plugin to use Logstash `data.path` subfolder.

It avoid to store state data into path that Logstash process doesn't have the rights to access.
After PR https://github.com/elastic/logstash/pull/12782, released with Logstash `7.13.0` , the owner of Logstash installation dir changed from `logstash` user to `root`, so the $HOME directory used by Logstash process is not  writable by it.
The process must store data only in the `data.path` folder. 
There's the expectation to create the smoothest path forward, so that users doesn't need to manually update their pipelines nor to manually intervene in moving files on their machines.

## Checklist

- ~~[] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files (and/or docker env variables)~~

## Author's Checklist

- [ ]

## Related issues

- 

## Use cases

As a user that uses the `record_last_run` I want to be able to upgrade to Logstash >= 7.13 without any throubles about access rights on internal storage files.

### DIFF
--- a/lib/logstash/filters/jdbc_static.rb
+++ b/lib/logstash/filters/jdbc_static.rb
@@ -152,9 +152,7 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
   public
 
   def register
-    # This check is Logstash 5 specific.  If the class does not exist, and it
-    # won't in older versions of Logstash, then we need to set it to nil.
-    @settings = defined?(LogStash::SETTINGS) ? LogStash::SETTINGS : nil
+    @settings = LogStash::SETTINGS
 
     prepare_data_dir
     prepare_runner
--- a/lib/logstash/filters/jdbc_static.rb
+++ b/lib/logstash/filters/jdbc_static.rb
@@ -189,7 +189,7 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
 
     # later, when local persistent databases are allowed set this property to LS_HOME/data/jdbc-static/
     # must take multi-pipelines into account and more than one config using the same jdbc-static settings
-    path_data = Pathname.new(LogStash::SETTINGS.get_value("path.data")).join("plugins", "filter", "jdbc_static")
+    path_data = Pathname.new(LogStash::SETTINGS.get_value("path.data")).join("plugins", "shared", "derby_home")
     path_data.mkpath
     java.lang.System.setProperty("derby.system.home", path_data.to_path)
     logger.info("derby.system.home is: #{java.lang.System.getProperty("derby.system.home")}")
--- a/lib/logstash/filters/jdbc_static.rb
+++ b/lib/logstash/filters/jdbc_static.rb
@@ -152,8 +152,6 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
   public
 
   def register
-    @settings = LogStash::SETTINGS
-
     prepare_data_dir
     prepare_runner
   end
@@ -183,15 +181,15 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
       begin
         ::File.delete()
       rescue Errno::EPERM => e
-        @logger.warn("Can't move file #{derby_log} due to access permissions")
+        @logger.warn("Can't delete file #{derby_log} due to access permissions")
       rescue e
-        @logger.warn("Can't move file #{derby_log}", {message => e.message})
+        @logger.warn("Can't delete file #{derby_log}", {message => e.message})
       end
     end
 
     # later, when local persistent databases are allowed set this property to LS_HOME/data/jdbc-static/
     # must take multi-pipelines into account and more than one config using the same jdbc-static settings
-    path_data = Pathname.new(@settings.get_value("path.data")).join("plugins", "filter", "jdbc_static")
+    path_data = Pathname.new(LogStash::SETTINGS.get_value("path.data")).join("plugins", "filter", "jdbc_static")
     path_data.mkpath
     java.lang.System.setProperty("derby.system.home", path_data.to_path)
     logger.info("derby.system.home is: #{java.lang.System.getProperty("derby.system.home")}")
--- a/lib/logstash/filters/jdbc_static.rb
+++ b/lib/logstash/filters/jdbc_static.rb
@@ -152,6 +152,10 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
   public
 
   def register
+    # This check is Logstash 5 specific.  If the class does not exist, and it
+    # won't in older versions of Logstash, then we need to set it to nil.
+    @settings = defined?(LogStash::SETTINGS) ? LogStash::SETTINGS : nil
+
     prepare_data_dir
     prepare_runner
   end
@@ -177,7 +181,7 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
   def prepare_data_dir
     # later, when local persistent databases are allowed set this property to LS_HOME/data/jdbc-static/
     # must take multi-pipelines into account and more than one config using the same jdbc-static settings
-    path_data = Pathname.new(settings.get_value("path.data")).join("plugins", "filter", "jdbc_static")
+    path_data = Pathname.new(@settings.get_value("path.data")).join("plugins", "filter", "jdbc_static")
     path_data.mkpath
     java.lang.System.setProperty("derby.system.home", path_data.to_path)
     logger.info("derby.system.home is: #{java.lang.System.getProperty("derby.system.home")}")
--- a/lib/logstash/filters/jdbc_static.rb
+++ b/lib/logstash/filters/jdbc_static.rb
@@ -177,6 +177,18 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
   private
 
   def prepare_data_dir
+    # cleanup existing Derby file left behind in $HOME
+    derby_log = "#{ENV['HOME']}/derby.log"
+    if (::File.exist?(derby_log))
+      begin
+        ::File.delete()
+      rescue Errno::EPERM => e
+        @logger.warn("Can't move file #{derby_log} due to access permissions")
+      rescue e
+        @logger.warn("Can't move file #{derby_log}", {message => e.message})
+      end
+    end
+
     # later, when local persistent databases are allowed set this property to LS_HOME/data/jdbc-static/
     # must take multi-pipelines into account and more than one config using the same jdbc-static settings
     path_data = Pathname.new(@settings.get_value("path.data")).join("plugins", "filter", "jdbc_static")
--- a/lib/logstash/filters/jdbc_static.rb
+++ b/lib/logstash/filters/jdbc_static.rb
@@ -179,11 +179,11 @@ module LogStash module Filters class JdbcStatic < LogStash::Filters::Base
     derby_log = "#{ENV['HOME']}/derby.log"
     if ::File.exist?(derby_log)
       begin
-        ::File.delete()
+        ::File.delete(derby_log)
       rescue Errno::EPERM => e
-        @logger.warn("Can't delete file #{derby_log} due to access permissions")
+        @logger.warn("Can't delete temporary file #{derby_log} due to access permissions")
       rescue e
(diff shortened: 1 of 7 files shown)


lib/logstash/inputs/jdbc.rb
```diff
@@ -230,11 +232,29 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
...
   attr_reader :database # for test mocking/stubbing
+  attr_reader :last_run_metadata_file_path # path to the file used as last run storage
 
   public
 
   def register
     @logger = self.logger
+
+    if record_last_run
+      if last_run_metadata_path.nil?
```
yaauie: In order to avoid this being a breaking transition, we need to make a one-time copy of the file in `#{ENV['HOME']}/.logstash_jdbc_last_run` to our new `path.data`-based default path. This will ensure that a user with an existing implicit configuration has continuity of behaviour before and after the upgrade.

andsel: It could be that after an upgrade with `deb`or `rpm` distribution the owner of the old folder change, so the old file can't be deleted.
The idea to implement here is:
-  if the file `#{ENV['HOME']}/.logstash_jdbc_last_run` exists and new one in `path.data` doesn't, then copy and try to delete the old. If the deletion fail, log a WARN informing the user he has to manually delete the old file.
- if exists both the old and new, print a WARN to inform the user and consider the new one to valid for use.

andsel: @yaauie I swtiched a little bit the flow described above:
if the `#{ENV['HOME']}/.logstash_jdbc_last_run` exists then move in the `path.data` sub-folder. If it can't because of access rights to the old file, then fail with a ConfigurationError so that the user has to intervene, fixing the access right to the old file or maybe moving it manually.

lib/logstash/filters/jdbc_static.rb
```diff
@@ -176,7 +180,9 @@ def loader_runner
   def prepare_data_dir
     # later, when local persistent databases are allowed set this property to LS_HOME/data/jdbc-static/
     # must take multi-pipelines into account and more than one config using the same jdbc-static settings
-    java.lang.System.setProperty("derby.system.home", ENV["HOME"])
+    path_data = Pathname.new(@settings.get_value("path.data")).join("plugins", "filter", "jdbc_static")
```
yaauie, line 183: `@settings` can be `nil` (per other comments, on LS5?), which will cause this line to throw a `NoMethodError`.

lib/logstash/inputs/jdbc.rb
```diff
@@ -230,11 +233,37 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
...
+  attr_reader :last_run_metadata_file_path # path to the file used as last run storage
 
   public
 
   def register
     @logger = self.logger
+
+    # This check is Logstash 5 specific.  If the class does not exist, and it
+    # won't in older versions of Logstash, then we need to set it to nil.
+    settings = defined?(LogStash::SETTINGS) ? LogStash::SETTINGS : nil
```
yaauie, line 245: similar issue here; LS5 is well past its end-of-life.

lib/logstash/inputs/jdbc.rb
```diff
@@ -230,11 +233,37 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
...
+
+    if @record_last_run
+      if @last_run_metadata_path.nil?
+        logstash_data_path = settings.get_value("path.data")
```
yaauie, line 249: `NoMethodError` sending `NilClass#get_value` on LS5

yaauie (dismissed):

lib/logstash/filters/jdbc_static.rb
```diff
@@ -151,6 +151,10 @@ def validate_value(value, validator)
   public
 
   def register
+    # This check is Logstash 5 specific.  If the class does not exist, and it
+    # won't in older versions of Logstash, then we need to set it to nil.
+    @settings = defined?(LogStash::SETTINGS) ? LogStash::SETTINGS : nil
```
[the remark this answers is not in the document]
andsel: The gemfile already impose that constraint so we can easily remove this LS 5 behavior
```
s.add_runtime_dependency "logstash-core", ">= 6.5.0"
```

lib/logstash/inputs/jdbc.rb
andsel: Fixed by c3fc4008165d65438928fc32f72208d68eaf6ae3

lib/logstash/inputs/jdbc.rb
andsel: Fixed by c3fc4008165d65438928fc32f72208d68eaf6ae3

lib/logstash/filters/jdbc_static.rb
andsel: Fixed by c3fc4008165d65438928fc32f72208d68eaf6ae3

lib/logstash/filters/jdbc_static.rb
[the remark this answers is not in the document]
andsel: Fixed by c3fc4008165d65438928fc32f72208d68eaf6ae3

lib/logstash/filters/jdbc_static.rb
```diff
@@ -176,7 +180,9 @@ def loader_runner
   def prepare_data_dir
     # later, when local persistent databases are allowed set this property to LS_HOME/data/jdbc-static/
     # must take multi-pipelines into account and more than one config using the same jdbc-static settings
-    java.lang.System.setProperty("derby.system.home", ENV["HOME"])
+    path_data = Pathname.new(@settings.get_value("path.data")).join("plugins", "filter", "jdbc_static")
+    path_data.mkpath
+    java.lang.System.setProperty("derby.system.home", path_data.to_path)
```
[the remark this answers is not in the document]
andsel: We don't have to copy safely these files, because on a restart, the `jdbc_static` filter, which uses an in-memory DB, has to recreate the Derby DB. The problem is that we left behind some trash. 
We have to clean it, but before have to find which files Derby creates.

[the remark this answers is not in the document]
andsel: It creates a file named "derby.log", which in a local test contains:
```
andrea@kalispera ~ $ cat derby.log 
----------------------------------------------------------------
Mon Mar 14 12:00:24 CET 2022:
Booting Derby version The Apache Software Foundation - Apache Derby - 10.14.1.0 - (1808820): instance a816c00e-017f-8816-1cb4-000051603af8 
on database directory memory:/home/andrea/c8f3ac67e8a5566f1a05a452 with class loader org.jruby.util.JRubyClassLoader@2da59753 
Loaded from file:/home/andrea/.m2/repository/org/apache/derby/derby/10.14.1.0/derby-10.14.1.0.jar
java.vendor=AdoptOpenJDK
java.runtime.version=11.0.11+9
user.dir=/home/andrea/workspace/logstash_configs/jdbc
os.name=Linux
os.arch=amd64
os.version=5.4.0-104-generic
derby.system.home=/home/andrea
Database Class Loader started - derby.database.classpath=''
```

Also described in https://db.apache.org/derby/docs/10.7/devguide/cdevdvlp27610.html.
We could remove it if we can, else warn the user we left a file behind to do a manual cleanup.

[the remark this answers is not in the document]
andsel: Fixed by b4ccbb8070185d2580c28fb9f91a5a0c42f7910d

andsel:
The failure on `7.x` and `8.x` SNAPSHOT is tracked in https://github.com/elastic/logstash/issues/13890

lib/logstash/inputs/jdbc.rb
```diff
@@ -374,4 +403,20 @@ def convert(column_name, value)
...
   end
+
+  def ensure_default_metadatafile_location(metadata_new_path)
+    old_default_path = Pathname.new("#{ENV['HOME']}/.logstash_jdbc_last_run")
+    if old_default_path.exist?
+      begin
+        # Previous versions of the plugin hosted the state file into $HOME/.logstash_jdbc_last_run.
+        # If there is a permission error in the move to the new location inform the user to give
+        # the correct access right before start
+        FileUtils.mv(old_default_path.to_path, metadata_new_path)
```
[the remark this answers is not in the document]
andsel: Fixed with 762c9045da48ca74317f30a225a4f1e519f7286e

spec/inputs/jdbc_spec.rb
```diff
@@ -1114,6 +1118,86 @@
...
+    let(:fake_home) do
+       path = Stud::Temporary.pathname
+       Pathname.new(path).tap {|path| path.mkpath}
+       path
+    end
+
+    context "when a file exists" do
+      before do
+        # in a faked HOME folder save a valid previous last_run metadata file
+        ENV['HOME'] = fake_home
```
yaauie, line 1184: nit: stubbing the `ENV['HOME']` ensures we don't pollute the `ENV` for subsequently-executed specs

```suggestion
        allow(ENV).to receive(:[]).with('HOME').and_return(fake_home)
```

yaauie (commented):
Changes and flow looks good. I've left a couple nitpicks.

andsel:
The :red_circle:  [CI](https://app.travis-ci.com/github/logstash-plugins/logstash-integration-jdbc/jobs/564398347) is due to a rubygems fix that land in `8.1.1`

lib/logstash/filters/jdbc_static.rb
```diff
@@ -175,9 +175,23 @@ def loader_runner
   private
 
   def prepare_data_dir
+    # cleanup existing Derby file left behind in $HOME
+    derby_log = "#{ENV['HOME']}/derby.log"
+    if ::File.exist?(derby_log)
+      begin
+        ::File.delete()
```
yaauie, line 182: We need to tell it which file to delete :)

```suggestion
        ::File.delete(derby_log)
```

yaauie (changes requested):
One change about specifying _which_ file to delete, and a couple nitpicks :)

lib/logstash/filters/jdbc_static.rb
andsel: :face_with_head_bandage:

lib/logstash/filters/jdbc_static.rb
```diff
@@ -175,9 +175,23 @@ def loader_runner
...
+        @logger.warn("Can't delete file #{derby_log} due to access permissions")
+      rescue e
+        @logger.warn("Can't delete file #{derby_log}", {message => e.message})
+      end
+    end
+
     # later, when local persistent databases are allowed set this property to LS_HOME/data/jdbc-static/
     # must take multi-pipelines into account and more than one config using the same jdbc-static settings
-    java.lang.System.setProperty("derby.system.home", ENV["HOME"])
+    path_data = Pathname.new(LogStash::SETTINGS.get_value("path.data")).join("plugins", "filter", "jdbc_static")
```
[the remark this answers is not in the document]
andsel: `derby.system.home` is a property used only by Derby and inside Logstash we use Derby  only in this plugin.
But I share the principle, do you think `plugins/derby_home` could be better?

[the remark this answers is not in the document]
yaauie: I'm +1 to either `plugins/shared/derby_home` or `plugins/derby_home`.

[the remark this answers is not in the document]
andsel: Gone with the `shared` one

andsel:
@yaauie the PR is ready for another round of review :-)

yaauie (approved):
LGTM 👍🏼

PR #109, opened 2022-05-02
Fix: do not execute more queries with debug logging

kares:
Parameters are already logged from Sequel, with `use_prepared_statement => true`:
```
[INFO ][logstash.inputs.jdbc     ][main][...] (0.005314s) PREPARE foo2: SELECT FIRST_NAME, LAST_NAME FROM employee WHERE EMP_NO >= ?
[INFO ][logstash.inputs.jdbc     ][main][...] (0.006059s) EXECUTE foo2; [2]
```
when `parameters` are set, these are clearly replaced (visible) in the query e.g. with `jdbc_paging_enabled => true`:
```
@statement="SELECT FIRST_NAME, LAST_NAME FROM employee WHERE EMP_NO = :p1", 
@jdbc_page_size=5, 
@parameters={:p1=>2}
```
```
[INFO ][logstash.inputs.jdbc     ][main][...] (0.008071s) SELECT CAST(current_setting('server_version_num') AS integer) AS v
[INFO ][logstash.inputs.jdbc     ][main][...] (0.000873s) SELECT count(*) AS "count" FROM (SELECT FIRST_NAME, LAST_NAME FROM employee WHERE EMP_NO = 2) AS "t1" LIMIT 1
[INFO ][logstash.inputs.jdbc     ][main][...] (0.000571s) SELECT * FROM (SELECT FIRST_NAME, LAST_NAME FROM employee WHERE EMP_NO = 2) AS "t1" LIMIT 5 OFFSET 0
```

---

finally, the added logging for the paging handler will log the total record count with every page:
```
[INFO ][logstash.inputs.jdbc     ] (0.001125s) SELECT CAST(current_setting('server_version_num') AS integer) AS v
[INFO ][logstash.inputs.jdbc     ] (0.001032s) SELECT count(*) AS "count" FROM (SELECT * FROM "employee" WHERE EMP_NO >= 0 ORDER BY EMP_NO) AS "t1" LIMIT 1
[DEBUG][logstash.inputs.jdbc     ] fetching paged dataset {:current_page=>1, :record_count=>1, :total_record_count=>2}
[INFO ][logstash.inputs.jdbc     ] (0.000584s) SELECT * FROM (SELECT * FROM "employee" WHERE EMP_NO >= 0 ORDER BY EMP_NO) AS "t1" LIMIT 1 OFFSET 0
[DEBUG][logstash.inputs.jdbc     ] fetching paged dataset {:current_page=>2, :record_count=>1, :total_record_count=>2}
[INFO ][logstash.inputs.jdbc     ] (0.000489s) SELECT * FROM (SELECT * FROM "employee" WHERE EMP_NO >= 0 ORDER BY EMP_NO) AS "t1" LIMIT 1 OFFSET 1

```
... so we have the same (count) information as before with debug logging, but it only happens for the paged case!

---
NOTE: instead of writing a specific spec I've added an integration test with paging, that turns on debug logging.


lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -16,27 +16,39 @@ def self.build_statement_handler(plugin, logger)
...
+
+    attr_reader :parameters
+
+    def initialize(plugin)
+      super(plugin)
+      @parameter_keys = ["sql_last_value"] + plugin.parameters.keys
+      @parameters = plugin.parameters.inject({}) do |hash,(k,v)|
+        case v
+        when LogStash::Timestamp
+          hash[k.to_sym] = v.time
```
andsel, line 44: `getTime` is deprecated  https://github.com/elastic/logstash/blob/e02a9e00c7b01c03db32bb737ebf4019fbb915ac/logstash-core/src/main/java/org/logstash/Timestamp.java#L85-L86 
Should we use `to_instant` ?

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -16,27 +16,39 @@ def self.build_statement_handler(plugin, logger)
...
   end
 
   class NormalStatementHandler < StatementHandler
+
+    attr_reader :parameters
+
+    def initialize(plugin)
+      super(plugin)
+      @parameter_keys = ["sql_last_value"] + plugin.parameters.keys
+      @parameters = plugin.parameters.inject({}) do |hash,(k,v)|
```
andsel, line 41: This loop modify the `parameters` converting `LogStash::Timestamp` parameters to something else (Joda Time in this case). Maybe extracting in a method would make the code more explicit.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -101,20 +110,29 @@ class ExplicitPagingModeStatementHandler < PagedNormalStatementHandler
     def perform_query(db, sql_last_value)
       query = build_query(db, sql_last_value)
       offset = 0
+      page_size = @jdbc_page_size
```
andsel, line 113: `@jdbc_page_size`is a not modifiable class field, it passed in constructor and never updated. I don't understand the need to copy into a local variable.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -52,27 +64,18 @@ def perform_query(db, sql_last_value)
 
     def build_query(db, sql_last_value)
       parameters[:sql_last_value] = sql_last_value
-      query = db[statement, parameters]
-      statement_logger.log_statement_parameters(statement, parameters, query)
-      query
+      db[statement, parameters]
```
andsel, line 67: Maybe I'm missing something, but in this case the original `statement_logger.log_statment_parameter` would ave logged the string:

```ruby
@logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters)
```

It's not clear, after removal of the statement logger, who logs the query here.

andsel (changes requested):
The idea seems good.
I've left some doubts about where the logging of the queries for `NormalStatmentHandler` is now moved.
Left a comment on a possible readability improvement, understand that's code moved from `post_init` method into `initialize`, would like to know your opinion, else it's ok how you moved.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
kares: there's no change really in terms of functionality - just a refactoring - [removed the `post_init` method](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/109/files#diff-18a21e2901a3bdc0983fab9b9c06cf6575f609433317586a8368e815009777e7L61-L71) and moved it in place into `initialize`. it was very confusing to read since there was `initialize` and `post_init` with a number of overrides.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
kares: yeah this is confusing but it pre-existed - just a refactoring of a `post_init` method: https://github.com/logstash-plugins/logstash-integration-jdbc/pull/109/files#diff-18a21e2901a3bdc0983fab9b9c06cf6575f609433317586a8368e815009777e7L62-L70

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
kares: the i-var won't be read in a loop repeatedly but most importantly it helped readability of the method since none of the other locals use the jdbc_ prefix and it looked a little "confusing" why one has the prefix but none of the others, now:
```ruby
        break unless rows_in_page == page_size
        offset += page_size
```
> it passed in constructor and never updated. I don't understand the need to copy into a local variable.

correct, again, mostly just refactored the `post_init` - it's a convention used through the whole file that plugin variables are copied to i-vars.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
kares: we already rely on query logging from Sequel and allow to customize the level at which SQL gets logged using `sql_log_level => ...` (`info` by default), the query logging already includes parameters the only missing piece is the total record count but that is only relevant when paging.

here's how it looked (before this PR) at debug level, notice the `Executing JDBC query` is redundant information and caused the extra count query:
```
[INFO ][logstash.inputs.jdbc     ][main] (0.000499s) SELECT count(*) AS "count" FROM (SELECT FIRST_NAME, LAST_NAME FROM employee WHERE EMP_NO = 2) AS "t1" LIMIT 1
[DEBUG][logstash.inputs.jdbc     ][main] Executing JDBC query {:statement=>"SELECT FIRST_NAME, LAST_NAME FROM employee WHERE EMP_NO = 2", :parameters=>{:sql_last_value=>2022-05-02 11:29:15 UTC}, :count=>1}
[INFO ][logstash.inputs.jdbc     ][main] (0.000528s) SELECT FIRST_NAME, LAST_NAME FROM employee WHERE EMP_NO = 2
```

PR #110, opened 2022-05-10
Refactor: start using scheduler mixin

kares:
#### part of https://github.com/elastic/logstash/issues/13773

see https://github.com/logstash-plugins/logstash-mixin-scheduler/pull/1


yaauie (approved):
:tada: fantastic! LGTM.

PR #122, opened 2023-02-14
Add an example of default_hash

dadoonet:
Could be useful to understand how `default_hash` works.

### DIFF
--- a/docs/filter-jdbc_static.asciidoc
+++ b/docs/filter-jdbc_static.asciidoc
@@ -111,10 +111,14 @@ filter {
         query => "SELECT firstname, lastname FROM users WHERE userid = ? AND country = ?"
         prepared_parameters => ["[loggedin_userid]", "[user_nation]"] <4>
         target => "user" <5>
+        default_hash => { <6>
+          firstname => nil
+          lastname => nil
+        }
       }
     ]
     # using add_field here to add & rename values to the event root
-    add_field => { server_name => "%{[server][0][description]}" } <6>
+    add_field => { server_name => "%{[server][0][description]}" } <7>
     add_field => { user_firstname => "%{[user][0][firstname]}" }
     add_field => { user_lastname => "%{[user][0][lastname]}" }
     remove_field => ["server", "user"]
@@ -139,7 +143,9 @@ See <<plugins-{type}s-{plugin}-object_order>>.
 follow the positional ordering.
 <5> Specifies the event field that will store the looked-up data. If the lookup
 returns multiple columns, the data is stored as a JSON object within the field.
-<6> Takes data from the JSON object and stores it in top-level event fields for
+<6> Creates a default user if the user can not be found in the users database. 
+In which case the <<plugins-{type}s-{plugin}-tag_on_default_use>> will be called.
+<7> Takes data from the JSON object and stores it in top-level event fields for
 easier analysis in Kibana.
 
 Here's a full example:


docs/filter-jdbc_static.asciidoc
```diff
@@ -139,7 +143,9 @@ See <<plugins-{type}s-{plugin}-object_order>>.
 follow the positional ordering.
 <5> Specifies the event field that will store the looked-up data. If the lookup
 returns multiple columns, the data is stored as a JSON object within the field.
-<6> Takes data from the JSON object and stores it in top-level event fields for
+<6> Creates a default user if the user can not be found in the users database.
```
andsel, line 146:
```suggestion
<6> Uses values for user fields if the user can not be found in the users database. 
```

docs/filter-jdbc_static.asciidoc
```diff
@@ -139,7 +143,9 @@ See <<plugins-{type}s-{plugin}-object_order>>.
 follow the positional ordering.
 <5> Specifies the event field that will store the looked-up data. If the lookup
 returns multiple columns, the data is stored as a JSON object within the field.
-<6> Takes data from the JSON object and stores it in top-level event fields for
+<6> Creates a default user if the user can not be found in the users database. 
+In which case the <<plugins-{type}s-{plugin}-tag_on_default_use>> will be called.
```
andsel, line 147:
```suggestion
In which case the <<plugins-{type}s-{plugin}-tag_on_default_use>> tags list is used to mark the event.
```

andsel (changes requested):
Left a couple of comments.

I would ask @karenzone  a quick check that we don't break docs, but the rest seems ok to me.

andsel:
@dadoonet if you add the `if` to check tag in outputs section would be great so that a user can immediately grasp how to leverage it.

dadoonet:
> if you add the `if` to check tag in outputs section would be great so that a user can immediately grasp how to leverage it.

I added it.

docs/filter-jdbc_static.asciidoc
```diff
@@ -139,7 +150,9 @@ See <<plugins-{type}s-{plugin}-object_order>>.
 follow the positional ordering.
 <5> Specifies the event field that will store the looked-up data. If the lookup
 returns multiple columns, the data is stored as a JSON object within the field.
-<6> Takes data from the JSON object and stores it in top-level event fields for
+<6> Uses values for user fields if the user can not be found in the users database. 
+In which case the <<plugins-{type}s-{plugin}-tag_on_default_use>> tags list is used to mark the event.
```
karenzone, line 154:
```suggestion
<6> Tags user fields with the format defined in <<plugins-{type}s-{plugin}-tag_on_default_use>> when a user is not found in the users database. 
```

karenzone (commented):
Left comment with alternate wording for your consideration.

docs/filter-jdbc_static.asciidoc
andsel: I think the meaning it's little bit different.
In case the user is not found in the database, then it still create a new event, but with the data that's provided into `default_hash`, which has some default values matching the columns.
So if a DB lookup result an empty set it brings the data from this "static default values" map, and tags the event with the list present in `tag_on_default_use`.

dadoonet: Ha right. I should not have commited the proposal... :(

dadoonet: I reverted my last commit

andsel: Expect @karenzone come up with a nice and fluent sentence to express that concept and avoid the previous tongue twister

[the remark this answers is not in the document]
andsel: To me this rephrase seems really good and smooth.

@karenzone 
>  Where does the "static default values" map get populated.

It's populate reading the `default_hash` setting in the [local_lookups](https://www.elastic.co/guide/en/logstash/8.6/plugins-filters-jdbc_static.html#plugins-filters-jdbc_static-local_lookups)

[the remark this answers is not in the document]
karenzone: I tweaked it again to provide more context to the `default_hash` setting to clearly explain the way those two settings work together.

dadoonet (commented):
LGTM

PR #119, opened 2022-10-28
Switch pattern of usage Sequel to leverage the DB connection pool

andsel:
## Release notes

Bugfix leak which happened in creating a new Database pool for every query. The pool is now crated on registration and closed on plugin's `stop`.

## What does this PR do?

This commit changes the way JDBC input plugin uses the Sequel DB access library. Instead of instantiating a new pool for every query, it creates the pool during registration of the plugin, then reuse that for every DB access. This avoid the connection leaked happening in scheduled queries, as described in #118.

In particular this PR does:
- moved the creation of Sequel Database from `execute_query()` method to plugin's `register`
- removed the reentrant lock that serialized access to the DB. Every DB connection access (creation, query & close) operations was guarded by this lock, which is not clear why existed due to the facts that `Database` is a connection pool per se and is thread safe.
- the close of the `Database` instance now happens only in the plugin's `#stop()` method, which was already present.

## Why is it important/What is the impact to the user?

With certain drivers, when using scheduled queries and reloading the pipeline, it produces a connection leak documented in #118.
The restart of the pipeline was important, because triggered the re-creation of internal plugin structures, like `PreparedStatementHandler` which is responsible for prepared statement re-assignmend accross multiple `Database` instances.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files (and/or docker env variables)~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~

## Author's Checklist

- [x] Run as described in the originating issue #118 

## How to test this PR locally

Follow the steps described in #118 

## Related issues

- Closes #118

PR #123, opened 2023-04-02
Added retry for sql statement execution failure

kaisecheng:
[jdbc-integration 5.4.1](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/119) has changed the way of using the connection pool. Since then, the sql statement execution does not retry for any failure. This PR adds a retry loop and uses the existing config `connection_retry_attempts` and `connection_retry_attempts_wait_time` to determine the number of retries and sleep time between retries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-## Unreleased
+## 5.4.2
   - Doc: described default_hash and tag_on_default_use interaction filter plugin [#122](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/122)
+  - Added retry for sql statement execution failure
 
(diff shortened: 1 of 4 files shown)


kaisecheng:
jenkins test this please

andsel (approved):
LGTM

tried shutting down and tearing up a local DB between the `statement_retry_attempts_wait_time` and worked as expected. Good work.

Maybe a nitpick, instead of `statement_retry_attempts_wait_time` would be better `statement_retry_attempts_interval`. I think that amongst the plugin's code base the `interval` suffix is more frequently used than `wait_time`. 
But feel free to skip the suggestion :-)

PR #125, opened 2023-04-18
Updates for JRuby 9.4

### LINKED ISSUE #133: Make the plugin compatible with JRuby 9.4 (Ruby 3.1)
Update the plugins code base so that it can be with JRuby 9.4.
The plugin should be proven to be installable on a Logstash instance that uses both JRuby 9.3 an 9.4.

andsel:
## Release notes
Fix: adaptations for JRuby 9.4

## What does this PR do?

This PR laid down adaptations to run this plugin under JRuby 9.4 (Ruby 3.1)

List of changes:
- in a method with var args the hash map has to be defined when passed as argument, d19fc58ab9d0eb1588811e9e43e78f80cefb8903  [Ruby 3.0](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/)
- Ruby 3.1 comes with Psych 4 which switched the alias of `load` from `unsafe_load` to `safe_load`  [reference](https://www.ctrl.blog/entry/ruby-psych4.html). Created a wrapper method to behave consistently across different versions of JRuby.
- fixed a stubbing error in tests: 72a2af524bd56f3d6bb22bfd744d4fb9a5d36a86

Let the user to use this plugin also on future Logstash versions which ship JRuby >= 9.4

## Checklist

- ~~[ ] I have commented my code, particularly in hard-to-understand areas~~
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files (and/or docker env variables)~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~

## Author's Checklist

- [x] run tests locally with JRuby 9.3 against a Logstash with JRuby 9.3
- [x] run tests locally with JRuby 9.4 against a Logstash with JRuby 9.4
- [x] run the plugin on a Logstash that ships JRuby 9.4

## Related issues

- Fixes #133 
- https://github.com/elastic/logstash/issues/14996

### DIFF
--- a/spec/plugin_mixins/jdbc/value_tracking_spec.rb
+++ b/spec/plugin_mixins/jdbc/value_tracking_spec.rb
@@ -1,0 +1,18 @@
+# encoding: utf-8
+require "logstash/plugin_mixins/jdbc/value_tracking"
+
+module LogStash module PluginMixins module Jdbc
+  describe ValueTracking do
+
+    let(:yaml_date_source) { "--- !ruby/object:DateTime '2023-06-15 09:59:30.558000000 +02:00'\n" }
+
+    context "#load_yaml" do
+      it "should load yaml with date string" do
+        parsed_date = LogStash::PluginMixins::Jdbc::ValueTracking.load_yaml(yaml_date_source)
+        expect(parsed_date.year).to eq 2023
+        expect(parsed_date.month).to eq 6
+        expect(parsed_date.day).to eq 15
+      end
+    end
+  end
+end end end
--- a/spec/plugin_mixins/jdbc/value_tracking_spec.rb
+++ b/spec/plugin_mixins/jdbc/value_tracking_spec.rb
@@ -3,15 +3,43 @@ require "logstash/plugin_mixins/jdbc/value_tracking"
 
 module LogStash module PluginMixins module Jdbc
   describe ValueTracking do
-
-    let(:yaml_date_source) { "--- !ruby/object:DateTime '2023-06-15 09:59:30.558000000 +02:00'\n" }
-
     context "#load_yaml" do
-      it "should load yaml with date string" do
-        parsed_date = LogStash::PluginMixins::Jdbc::ValueTracking.load_yaml(yaml_date_source)
-        expect(parsed_date.year).to eq 2023
-        expect(parsed_date.month).to eq 6
-        expect(parsed_date.day).to eq 15
+
+      context "with date string" do
+        let(:yaml_date_source) { "--- !ruby/object:DateTime '2023-06-15 09:59:30.558000000 +02:00'\n" }
+
+        it "should be loaded" do
+          parsed_date = LogStash::PluginMixins::Jdbc::ValueTracking.load_yaml(yaml_date_source)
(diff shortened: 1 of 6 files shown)


robbavey (changes requested):
Code changes look good to me, but can we please add unit tests for the value tracker?

andsel:
@robbavey thank't for reviewing, integrated your suggestion with commit cf81c3e03213ca7504b7a6525add92e44a10dee7

spec/plugin_mixins/jdbc/value_tracking_spec.rb
```diff
@@ -0,0 +1,18 @@
+# encoding: utf-8
+require "logstash/plugin_mixins/jdbc/value_tracking"
+
+module LogStash module PluginMixins module Jdbc
+  describe ValueTracking do
+
+    let(:yaml_date_source) { "--- !ruby/object:DateTime '2023-06-15 09:59:30.558000000 +02:00'\n" }
+
+    context "#load_yaml" do
```
robbavey, line 9: Can we have tests that verify a few different scenarios - `Time`, `DateTime`, `BigDecimal`? I also wonder if we should have a test that uses the `build_last_value_tracker(plugin)` to do `read` and `write` these values too

robbavey (changes requested):
Thank you for adding a test! Can we add a couple more to make sure we cover read/write across all scenarios?

spec/plugin_mixins/jdbc/value_tracking_spec.rb
andsel: Yep, however I think that testing  `build_last_value_tracker` is a little bit out of context of this PR. We can avoid to push 2 PRs, one for JRuby 9.4 updates and another to cover with test existing code, and commit everything to this

robbavey: I'm ok with that

andsel: Done:
- with commit 743f919279f11048ceaa6865235ab8cc50a356bf `#load_yaml` has been covered for Time and BigDecimal cases
- while with 9a640a724c0dc695d0ddca6f2ef69f367d6b90ea the method `#build_last_value_tracker` has been exercised

spec/plugin_mixins/jdbc/value_tracking_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+          allow(plugin).to receive(:jdbc_default_timezone).and_return(:something_not_nil)
+        end
+
+        it "should write correctly" do
+          tracker = ValueTracking.build_last_value_tracker(plugin)
+          tracker.set_value("2023-06-15T15:28:15+02:00")
+          tracker.write
+
+          temp_file.rewind
+          v = ValueTracking.load_yaml(::File.read(temp_file.path))
```
robbavey, line 87: Can we use `tracker.read` here? (Or create a separate reader tracker, and use `reader_tracker.read`?)

Issue #118, opened 2022-10-18
Connection leak on pipeline reload when using Oracle driver


andsel:
**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`) `7.17.3` but was verified also on `7.17.7` and `8.4.3`
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker) `tar.gz`
4. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes) `bin/logstash`
5. How was the Logstash Plugin installed: already present

**JVM** (e.g. `java -version`):

If the affected version of Logstash is 7.9 (or earlier), or if it is NOT using the bundled JDK or using the 'no-jdk' version in 7.10 (or higher), please provide the following information:

1. JVM version (`java -version`) the bundled one
2. JVM installation source (e.g. from the Operating System's package manager, from source, etc). bundled
3. Value of the `JAVA_HOME` environment variable if set.

**OS version** (`uname -a` if on a Unix-like system):
```
Linux kalimera 5.15.0-50-generic #56~20.04.1-Ubuntu SMP Tue Sep 27 15:51:29 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

**Description of the problem including expected versus actual behavior**:
When the `logstash-input-jdbc` plugin is used to pull data out of an Oracle DB `21.3.0.0` but also `19.x`, and the pipeline is reloaded, then listing the open connections with `lsof` the number increases at every run.

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including (e.g.) pipeline definition(s), settings, locale, etc.  The easier
you make for us to reproduce it, the more likely that somebody will take the
time to look at it.

 1. download and install  docker image for Oracle Enterprise server from https://container-registry.oracle.com/ords/f?p=113:4:110741341869369.
   1.1. create an account in the web interface of the Oracle site.
   1.2. docker login and start it
   ```bash
   docker login container-registry.oracle.com
docker pull container-registry.oracle.com/database/enterprise:21.3.0.0
[...]

PR #130, opened 2023-05-11
Session handling reprise

yaauie:
Resolves the [described connection leak in prepared statements](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/118#issuecomment-1290951734) in an alternate way to the solution presented in #119, restoring the prior behavior for the input to make one DB connection per scheduler invocation.

 - in v5.4.0 and before, prepared statement handler leaked connections
 - in v5.4.1 connection moved from run-phase to register, which causes connection issues to prevent pipeline from starting; this is problematic for customers whose pipeline reloads fail because it requires human intervention, and also causes undesired long-lived connections.
 - this PR reverts the v5.4.1 fix, and solves the original issue by refactoring the prepared statement handler to avoid the connection leak

_However_:

 - v5.4.1-fix had an accidentally-desired behavior change, in which unscheduled instances that failed to connect would fail in `register`, which would crash the pipeline.
 - Prior to that fix, failing to connect would fail in `execute_statement`, which would only lead to a log message; when the job was unscheduled, this would lead to the plugin closing normally

Question:

 - when the plugin has no schedule, and cannot connect at startup, is it desirable to continue to crash the plugin (and therefore the pipeline)?

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -261,6 +261,8 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
       end
     end
 
+    prepare_jdbc_connection
+
     if @use_column_value
       # Raise an error if @use_column_value is true, but no @tracking_column is set
       if @tracking_column.nil?
@@ -303,20 +305,6 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
         converters[encoding] = converter
       end
     end
-
-    load_driver
-    begin
-      open_jdbc_connection
-    rescue Sequel::DatabaseConnectionError,
-           Sequel::DatabaseError,
-           Sequel::InvalidValue,
-           Java::JavaSql::SQLException => e
-      details = { exception: e.class, message: e.message }
-      details[:cause] = e.cause.inspect if e.cause
-      details[:backtrace] = e.backtrace if @logger.debug?
-      @logger.warn("Exception when executing JDBC query", details)
-      raise(LogStash::ConfigurationError, "Can't create a connection pool to the database")
-    end
   end # def register
 
   # test injection points
@@ -329,6 +317,7 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
   end
 
   def run(queue)
+    load_driver
     if @schedule
       # scheduler input thread name example: "[my-oracle]|input|jdbc|scheduler"
       scheduler.cron(@schedule) { execute_query(queue) }
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -305,6 +305,7 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
         converters[encoding] = converter
       end
     end
+    load_driver
   end # def register
 
   # test injection points
@@ -317,7 +318,6 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
   end
 
   def run(queue)
-    load_driver
     if @schedule
       # scheduler input thread name example: "[my-oracle]|input|jdbc|scheduler"
       scheduler.cron(@schedule) { execute_query(queue) }
--- a/lib/logstash/plugin_mixins/jdbc/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc/jdbc.rb
@@ -196,13 +196,21 @@ module LogStash  module PluginMixins module Jdbc
     end
 
     public
+    def prepare_jdbc_connection
+      @connection_lock = ReentrantLock.new
+    end
+
+    public
     def close_jdbc_connection
       begin
         # pipeline restarts can also close the jdbc connection, block until the current executing statement is finished to avoid leaking connections
         # connections in use won't really get closed
+        @connection_lock.lock
         @database.disconnect if @database
       rescue => e
         @logger.warn("Failed to close connection", :exception => e)
+      ensure
+        @connection_lock.unlock
       end
     end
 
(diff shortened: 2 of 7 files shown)


yaauie:
This clashes with the subsequent remediations introduced in #123 🤦🏼 

I'll assess how to assimilate the two shortly.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -148,41 +148,28 @@ def perform_query(db, sql_last_value)
     private
 
     def build_query(db, sql_last_value)
-      @parameters = create_bind_values_hash
-      if statement_prepared.false?
-        prepended = parameters.keys.map{|v| v.to_s.prepend("$").to_sym}
-        @prepared = db[statement, *prepended].prepare(:select, name)
```
yaauie, line 154: review note: this `@prepared` ivar holds a reference to the `db` that was used to create it, which made it a source of connection leaks.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -148,41 +148,28 @@ def perform_query(db, sql_last_value)
...
-      begin
-        db.call(name, parameters)
-      rescue => e
-        # clear the statement prepared flag - the statement may be closed by this
-        # time.
-        statement_prepared.make_false
-        raise e
-      end
+      prepared = db.prepared_statement(name)
+      prepared ||= db[statement, *positional_bind_placeholders].prepare(:select, name)
```
yaauie, line 154: review note: Sequel's `Dataset#prepare` always registers the resulting prepared statement with the `Database` so that it can be looked up with `Database#prepared_statement(name)`.

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -148,41 +148,28 @@ def perform_query(db, sql_last_value)
...
+
+      prepared.call(positional_bind_mapping(sql_last_value))
```
yaauie, line 156: review note: Sequel's `Database#call(name, parameters)` is a convoluted way of looking up the prepared statement by name and then sending it `PreparedStatement#call(parameters)`.

yaauie (commented):
There are two commits here; a revert of the 5.4.1 fix, plus a clean alternate fix. It is best reviewed one commit at a time.

lib/logstash/plugin_mixins/jdbc/jdbc.rb
```diff
@@ -195,14 +198,22 @@ def open_jdbc_connection
       end
     end
 
+    public
+    def prepare_jdbc_connection
+      @connection_lock = ReentrantLock.new
```
donoghuc, line 203: Is there a reason to use `ReentrantLock` instead of `Mutex`? I think with `Mutex` we could use its `synchronize` block and not have to explicitly handle unlocking with `ensure`.

donoghuc (commented):
Overall I love the solution. I assume stopping re-using connections wont be too expensive. I'm for sure on board with this. One question about using the automatic cleanup of `Mutex` vs `ReentrantLock` for synchronizing but other than that i think this is great.

lib/logstash/plugin_mixins/jdbc/jdbc.rb
yaauie: This comes from the reverted commit, and is just returning what was there before, but I agree with you. The `ReentrantLock#synchronize` would still be useful for clarity. I'd be glad to refactor.

donoghuc: oh, wow. Yeah that is absolutely an issue. Nice analysis.

lib/logstash/plugin_mixins/jdbc/jdbc.rb
```diff
@@ -159,82 +161,94 @@ def log_java_exception(e)
...
-      rescue => e
-        @logger.warn("Failed to close connection", :exception => e)
       end
+    rescue => e
+      @logger.warn("Failed to close connection", :exception => e)
     end
 
     public
-    def execute_statement
+    def execute_statement(&result_handler)
```
donoghuc, line 219:
```suggestion
    def execute_statement(&_result_handler)
```
Nit: I like explicitly calling out the expected block in the signature, but the underscore helps indicate we don't reference it explicitly.

donoghuc (approved):
One small question https://github.com/logstash-plugins/logstash-integration-jdbc/pull/130#discussion_r1892802921 and style suggestion https://github.com/logstash-plugins/logstash-integration-jdbc/pull/130#discussion_r1892848731

The refactor to use `synchronize` looks much better. I'm not sure what the normal use cases are around query frequency... Assuming forcing new connections on every run will be a noticeable performance change?

lib/logstash/inputs/jdbc.rb
```diff
@@ -329,6 +317,7 @@ def set_value_tracker(instance)
   end
 
   def run(queue)
+    load_driver
```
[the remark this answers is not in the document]
yaauie: This is another part of the reverted commit, and I'm not entirely sure why it was that way before.

I _do_ know that failures in `#register` will prevent pipeline instantiation, while failures in `#run` will occur after the pipeline has been initialized, and that it would be better to continue aborting early if we cannot load the driver. 

I'll make a stab at bouncing it back up into `#register`, and see what ramifications that has on tests.

yaauie (commented):
> Assuming forcing new connections on every run will be a noticeable performance change?

The most rapidly this can be scheduled is once per minute, and we have seen issues where people are frustrated that their daily jobs are keeping connections alive since the v5.4.1 change in behaviour. 

We may want to introduce connection keepalive at some point for those who want it, but I believe that should be done separately if at all.

PR #145, opened 2023-11-29
Added an assertion to a test to verify the file effectively exists and is read only

andsel:


mashhurs (approved):
LGTM!

Issue #150, opened 2024-01-18
setting jdbc_default_timezone in one plugin instance affects behavior of plugins where it is not set


yaauie:
## Logstash information:

Please include the following information:

1. Logstash version: 7.x -> 8.x
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker): any
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes): any
4. How was the Logstash Plugin installed: default

**JVM** (e.g. `java -version`): any

**OS version** (`uname -a` if on a Unix-like system): any

## Description of the problem including expected versus actual behavior**:

 - GIVEN: a database table with times stored in a local timezone
 - GIVEN: a logstash instance in the _same_ shared local timezone
 - OBSERVED: an instance of the JDBC input plugin configured _without_ specifying `jdbc_default_timezone` will load timestamps correctly (that is: `2023-12-31T18:00:00.000` in `America/Chicago` becomes `2024-01-01T00:00:00.000Z`)

 - HOWEVER: if another instance of the jdbc input plugin whose `jdbc_default_timezone` is set exists (or has ever existed) in the process:
 - OBSERVED: an instance of the JDBC input plugin configured _without_ specifying `jdbc_default_timezone` will load timestamps incorrectly (that is: `2023-12-31T18:00:00.000` in `America/Chicago` becomes `2023-12-31T18:00:00.000Z`)

## Workaround

_Always_ supply the `jdbc_default_timezone` when defining jdbc input plugins, optionally using  `TZ` environment variable that is present on most systems and controls the Logstash process's timezone:

~~~
    jdbc_default_timezone => "${TZ}"
~~~

## Steps to reproduce:

This is easiest to reproduce in postgresql, although the concept applies to _any_ database vendor.

0. download the jdbc driver jar, and export its absolute path to `POSTGRESQL_JDBC_DRIVER_JAR`:
   ~~~
   wget https://jdbc.postgresql.org/download/postgresql-42.7.1.jar
   export POSTGRESQL_JDBC_DRIVER_JAR="${PWD}/postgresql-42.7.1.jar"
   ~~~
2. bootstrap a local postgresql server and load it with data. In this case we create a _view_ containing rows that have equivalent timestamps from the perspective of different time zones:
   ~~~
   alter database postgres set timezone to 'America/Chicago';
[...]

Issue #107, opened 2022-03-15
Fails on timestamp comparison


andsel:
**Build scan**:
Issue seen locally.

**Repro line**:
Set the local timezone to something different than UTC.

- this fails
```
TZ="Europe/Rome" bundle exec rspec -fd spec/inputs/jdbc_spec.rb:1469
```
- this successed
```
TZ=UTC bundle exec rspec -fd spec/inputs/jdbc_spec.rb:1469
```

```
bundle exec rspec spec/inputs/jdbc_spec.rb
```
or 
```
bundle exec rspec -fd spec/inputs/jdbc_spec.rb:1469
```

**Reproduces locally?**:
yes

**Applicable branches**:
`main` commit 6a75566cad54805ff05a50a8a4879aede54a7967

**Failure history**:

**Failure excerpt**:
```
Failures:

  1) LogStash::Inputs::Jdbc when fetching Various Typed data should convert all columns to valid Event acceptable data types
     Failure/Error: expect(event.get("started_at")).to be_a_logstash_timestamp_equivalent_to("1999-12-31T00:00:00.000Z")
       expected 1999-12-30T23:00:00Z to be a LogStash::Timestamp equivalent to 1999-12-31T00:00:00Z
     # ./spec/inputs/jdbc_spec.rb:1429:in `block in <main>'
     # /home/andrea/.rvm/gems/jruby-9.2.19.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in <main>'
     # /home/andrea/.rvm/gems/jruby-9.2.19.0/gems/logstash-devutils-2.4.0-java/lib/logstash/devutils/rspec/spec_helper.rb:61:in `block in <main>'

  2) LogStash::Inputs::Jdbc when debug logging and a count query raises a count related error if the count query raises an error should create an event normally
     Failure/Error: expect(event.get("started_at")).to be_a_logstash_timestamp_equivalent_to("1999-12-31T00:00:00.000Z")
       expected 1999-12-30T23:00:00Z to be a LogStash::Timestamp equivalent to 1999-12-31T00:00:00Z
     # ./spec/inputs/jdbc_spec.rb:1476:in `block in <main>'
     # /home/andrea/.rvm/gems/jruby-9.2.19.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in <main>'
[...]

Issue #53, opened 2020-11-19
plugin instance's timezone affects other plugins


kares:
When a plugin is setup to connect to a data-source (such as Oracle) timestamps are assumed UTC.

However, when a second plugin is configured and sets: `'jdbc_default_timezone' => "Etc/GMT+1"`
this ends up [having](https://github.com/logstash-plugins/logstash-input-jdbc/blob/ae2523b5810eca7c5aa0a19b8677b49b643318fb/lib/logstash/plugin_mixins/jdbc/jdbc.rb#L187-L188) [consequences](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/extensions/named_timezones.rb#L52) for the first plugin's behavior. 

This can be traced all the way down to adjusting `Sequel.datetime_class`.

Let's suppose Oracle contains the following tracking_column (of type `timestamp`): 
`TO_TIMESTAMP('31-01-2020 20:30:40.123456', 'dd-mm-yyyy HH24:MI:SS.FF'))` 
(local timezone is GMT+1 - no `ENV['TZ']` set)

1. by default (no `jdbc_detault_timezone` set) the plugin will roll with `Sequel.datetime_class = Time` : 
    Sequel reads the timestamp as `:updated_at=>2020-01-31 19:30:40 UTC`
    last run value is stored as `2020-01-31 19:30:40 UTC` (`Time`)

2. adding an explicit `Sequel.datetime_class = DateTime` changes behavior to:
   timestamp value: `#<DateTime: 2020-01-31T20:30:40+00:00 ((2458880j,73840s,123456000n),+0s,2299161j)>`
  last run value is stored as `2020-01-31 20:30:40 +0000` (`Time`)

let's suppose the next row contains a +1 minute timestamp: 
`TO_TIMESTAMP('31-01-2020 20:31:40','dd-mm-yyyy HH24:MI:SS'))`

while case (2) seems more correct the plugin's behavior will be in case (1) read the next row with the above timestamp while in case (2) the row is going to be skipped (assumed already read).

(1): `WHERE updated_at > TIMESTAMP '2020-01-31 19:30:40.123456 +00:00' ORDER BY updated_at` :green_circle:  
    (next row's timestamp is assumed to be `2020-01-31 19:32:40 UTC`)

(2): `WHERE updated_at > TIMESTAMP '2020-01-31 20:30:40.123456 +00:00' ORDER BY updated_at` :red_circle: 

[...]

PR #151, opened 2024-01-23
Sequel datetime class consistency

yaauie:
Resolves a bug in which any one instance of a JDBC input plugin using `jdbc_default_timezone` changes the behaviour of plugin instances that do _not_ use `jdbc_default_timezone`, by:
 - (a) _always_ loading Sequel's `named_timezones` extension, and
 - (b) preventing `Sequel#datetime_class` from being anything other than its original default value.

This PR has two commits:
 - one with specs that _succeed_ in isolation but _fail_ when run with the rest of the spec suite (see commit message for details)
 - one with code-changes that ensure the above-provided specs _always_ succeed, and that this plugin _always_ uses the local context to parse times absent explicit configuration for the individual plugin, regardless of how other plugins are configured.

Resolves: #150 
Relates: #53 
Relat3es: #107


andsel (approved):
LGTM

Issue #152, opened 2024-01-30
Logstash does not reconnect when Oracle "SQLRecoverableException" is raised


edmocosta:
This problem is caused when a connection gets closed by Oracle (idle sessions, workload rules, network issues, etc.)
Logstash should gracefully recover from those errors, discarding the stale connection and reconnecting the client. Instead, the plugin hangs trying to reuse the closed/invalid connections.

It's the same `sequel` DB2 issue https://github.com/logstash-plugins/logstash-integration-jdbc/issues/143, but this time for oracle databases.  

Fortunately, the latest `sequel` version is already fixed, so bumping the library should solve the problem.

PR #153, opened 2024-01-30
Update sequel version to >= 5.74.0

edmocosta:
Updated the `sequel` gem version to >= [5.74.0](https://sequel.jeremyevans.net/rdoc/files/doc/release_notes/5_74_0_txt.html)

---
Closes: https://github.com/logstash-plugins/logstash-integration-jdbc/issues/152

### DIFF
--- a/logstash-integration-jdbc.gemspec
+++ b/logstash-integration-jdbc.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name = 'logstash-integration-jdbc'
-  s.version         = '5.4.6'
+  s.version         = '5.4.7'
   s.licenses = ['Apache License (2.0)']
   s.summary         = "Integration with JDBC - input and filter plugins"
   s.description     = "This gem is a Logstash plugin required to be installed on top of the Logstash core pipeline using $LS_HOME/bin/logstash-plugin install gemname. This gem is not a stand-alone program"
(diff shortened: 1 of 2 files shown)


logstash-integration-jdbc.gemspec
```diff
@@ -29,7 +29,9 @@ Gem::Specification.new do |s|
   # Restrict use of this plugin to versions of Logstash where support for integration plugins is present.
   s.add_runtime_dependency "logstash-core", ">= 6.5.0"
   s.add_runtime_dependency 'logstash-codec-plain'
-  s.add_runtime_dependency 'sequel', '>= 5.73.0'
+  # Pinned sequel version to >= 5.75.0 as it fixes the generic jdbc adapter to properly
+  # handle disconnection errors, and avoid stale connections in the pool.
+  s.add_runtime_dependency 'sequel', '>= 5.75.0'
```
jsvd, line 34:
```suggestion
  # Pinned sequel version to >= 5.74.0 as it fixes the generic jdbc adapter to properly
  # handle disconnection errors, and avoid stale connections in the pool.
  s.add_runtime_dependency 'sequel', '>= 5.74.0'
```

Given the fix was introduced with https://github.com/jeremyevans/sequel/commit/813db795a77687b9a639fe889b038e7a0e46829e and shipped with 5.74.0, maybe we can change this to 5.74.0 instead?

edmocosta: Agreed, it makes sense! Thanks!

jsvd (approved):
LGTM

PR #165, opened 2024-05-07
fix: use `Array#<<` to append an element to an array

yaauie:
Ruby's `Array#concat` needs an _Array_ argument; appending a single element can be done with `Array#<<`.

Fixes exception that could occur repeatedly when the upstream loader is not currently connected:

~~~
TypeError, original message: no 
implicit conversion of String into Array", :backtrace=>["org/jruby/RubyArray.java:1670:in `concat'", 
"/usr/share/logstash/vendor/bundle/jruby/2.6.0/gems/logstash-integration-jdbc-
5.4.1/lib/logstash/filters/jdbc/read_only_database.rb:15:in `count'",
~~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
+## 5.4.11
+- Fixes an exception that could occur while reloading `jdbc_static` databases when the underlying connection to the remote has been broken [#165](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/165)
+
 ## 5.4.10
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
-## Unreleased
- - Adds retry mechanism when checkout Derby from SVN repository [#158](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/158)
```
yaauie, line 2: I removed this from the changelog because it is build-chain related change that is not user-facing or part of the released artifacts.

andsel (approved):
LGTM

PR #168, opened 2024-05-15
Upgrade Gradle to 8.7

andsel:
Update Gradle to 8.7.


jsvd (changes requested):
requires upgrading gradle wrapper jar and scripts

jsvd (approved):
LGTM

PR #170, opened 2024-05-21
Add support for "uuid" type in tracking_column_type

Nyamops:
* Added "uuid" type to tracking_column_type settings
* Updated relevant documentation
* Added tests for "uuid" type
* Ensured compatibility with existing types

### DIFF
--- a/docs/input-jdbc.asciidoc
+++ b/docs/input-jdbc.asciidoc
@@ -127,7 +127,7 @@ Here is the list:
 
 |==========================================================
 |sql_last_value | The value used to calculate which rows to query. Before any query is run,
-this is set to Thursday, 1 January 1970, or 0 if `use_column_value` is true and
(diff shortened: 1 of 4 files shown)


m-khvoinitsky:
Why? Considering that UUIDs aren’t ordered.

Nyamops:
> Why? Considering that UUIDs aren’t ordered.

UUID v1-v4 are indeed unordered. UUID v7 was introduced in the 2022 [ietf draft](https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/00/) and now formally standardized in [rfc 9562](https://datatracker.ietf.org/doc/rfc9562/):

> UUIDv7 features a time-ordered value field derived from the widely
   implemented and well-known Unix Epoch timestamp source, the number of
   milliseconds since midnight 1 Jan 1970 UTC, leap seconds excluded.

Issue #126, opened 2023-04-18
SQL-Query that works with input-jdbc plugin doesn't work with jdbc_streaming filter plugin


s137:
**Logstash information**:

1. Logstash version: 8.3.3
2. Logstash installation source: extracted from the official zip archive
3. How is Logstash being run: windows service or console
4. How was the Logstash Plugin installed: plugin was not installed separately
5. Plugin Version: 5.3.0

**JVM** (e.g. `java -version`): Bundled JDK:
openjdk version "11.0.15" 2022-04-19
OpenJDK Runtime Environment Temurin-11.0.15+10 (build 11.0.15+10)
OpenJDK 64-Bit Server VM Temurin-11.0.15+10 (build 11.0.15+10, mixed mode)

**OS version**: Windows 10

**Description of the problem including expected versus actual behavior**:

The following SQL-Statement works fine with the jdbc-input plugin, but fails to execute with the jdbc_streaming filter plugin:

`SELECT EditDate FROM Test` 

The EditDate here would be of the type "smalldatetime" (MSSQL-Server).

When trying to run the statment via the jdbc_streaming filter plugin, the following error occurs and breaks the whole pipeline:

````
[...]

PR #171, opened 2024-08-08
Add support for date columns to jdbc static and streaming filters

edmocosta:
This PR adds support for SQL `DATE` columns to the `jdbc` static and streaming filters, similar to the existing input plugin support.

Before this changes, querying data from tables with `DATE`  columns would produce the error `Missing Converter handling for full class name=org.jruby.ext.date.RubyDateTime, simple name=RubyDateTime`. 

---
Closes: https://github.com/logstash-plugins/logstash-integration-jdbc/issues/126

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/value_handler.rb
+++ b/lib/logstash/plugin_mixins/jdbc/value_handler.rb
@@ -1,0 +1,26 @@
+# encoding: utf-8
+require "time"
+require "date"
+
+module LogStash module PluginMixins module Jdbc
+  # Provides functions to extract the row's values, ensuring column types
+  # are properly decorated to become coercible to a LogStash::Event.
+  module ValueHandler
+    # Stringify the row keys and decorate values when necessary
+    def extract_values_from(row)
+      Hash[row.map { |k, v| [k.to_s, decorate_value(v)] }]
+    end
+
+    # Decorate the value so it can be used as a LogStash::Event field
+    def decorate_value(value)
+      case value
+      when Time
+        LogStash::Timestamp.new(value)
+      when Date, DateTime
+        LogStash::Timestamp.new(value.to_time)
+      else
+        value
+      end
+    end
+  end
+end end end
--- a/lib/logstash/plugin_mixins/jdbc/value_handler.rb
+++ b/lib/logstash/plugin_mixins/jdbc/value_handler.rb
@@ -14,10 +14,8 @@ module LogStash module PluginMixins module Jdbc
     # Decorate the value so it can be used as a LogStash::Event field
     def decorate_value(value)
       case value
-      when Time
-        LogStash::Timestamp.new(value)
       when Date, DateTime
-        LogStash::Timestamp.new(value.to_time)
+        value.to_time
       else
         value
       end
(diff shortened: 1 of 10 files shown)


lib/logstash/plugin_mixins/jdbc/value_handler.rb
```diff
@@ -0,0 +1,26 @@
...
+    # Decorate the value so it can be used as a LogStash::Event field
+    def decorate_value(value)
+      case value
+      when Time
+        LogStash::Timestamp.new(value)
+      when Date, DateTime
+        LogStash::Timestamp.new(value.to_time)
+      else
+        value
+      end
```
jsvd, line 23: Do all tests still pass like this?

```suggestion
      case value
      when Date, DateTime
        value.to_time
      else
        value
      end
```

edmocosta: Yes, they still pass with the suggested code. It seems those conversions are not needed anymore, they were probably added [here](https://github.com/logstash-plugins/logstash-integration-jdbc/blob/main/lib/logstash/plugin_mixins/jdbc/jdbc.rb#L263) for previous LS versions  `<= 6.5.0`. Should I change to this new code?

jsvd: Althought performance wise it won't make a difference as the object should be the same, I do suggest we have simpler since the Valuefier already handles Time objects both in 8.x and 7.x

spec/filters/integration/jdbc_static_spec.rb
```diff
@@ -116,6 +116,39 @@ module LogStash module Filters
...
+
+            now = DateTime.now
+            sever = event.get("server").first
+            expect(sever['ip']).to eq("10.3.1.1")
+            expect(sever['name']).to eq("mv-server-1")
+            expect(sever['location']).to eq("MV-9-6-4")
+            expect(sever['entry_date']).to eq(LogStash::Timestamp.new(Time.new(2003, 2, 1)))
+            expect(sever['entry_time']).to eq(LogStash::Timestamp.new(Time.new(now.year, now.month, now.day, 10, 5, 0)))
+            expect(sever['timestamp']).to eq(LogStash::Timestamp.new(Time.new(2003, 2, 1, 1, 2, 3)))
+          end
```
jsvd, line 150: I'd prefer if we have separate tests for each data type, in case one fails it's easier to read the rspec output:

```suggestion
          before(:each) { plugin.register }
          subject { event.get("server").first }
          it "maps the DATE to a Logstash Timestamp" do
             plugin.filter(event)
             expect(subject['entry_date']).to eq(LogStash::Timestamp.new(Time.new(2003, 2, 1)))
          end

          it "maps the TIME field to a Logstash Timestamp" do
            plugin.filter(event)
            now = DateTime.now
            expect(subject['entry_time']).to eq(LogStash::Timestamp.new(Time.new(now.year, now.month, now.day, 10, 5, 0)))
          end

           it "maps the TIMESTAMP to a Logstash Timestamp" do
             plugin.filter(event)
             expect(subject['timestamp']).to eq(LogStash::Timestamp.new(Time.new(2003, 2, 1, 1, 2, 3)))
          end
```

jsvd (approved):
LGTM :shipit:

Issue #154, opened 2024-01-30
Missing documentation for `statement_retry_attempts` and `statement_retry_attempts_wait_time` options


edmocosta:
#### Tell us about the issue

**Description:**  
There's no documentation for the JDBC input plugin `statement_retry_attempts` and `statement_retry_attempts_wait_time` options. It should have similar entries on the settings table and descriptions.

**URL:** 
https://www.elastic.co/guide/en/logstash/8.12/plugins-inputs-jdbc.html#plugins-inputs-jdbc-options

PR #177, opened 2024-12-03
Document `statement_retry_attempts` and `statement_retry_attempts_wait_time` options

jsvd:
closes https://github.com/logstash-plugins/logstash-integration-jdbc/issues/154


donoghuc (approved):
Matches source https://github.com/logstash-plugins/logstash-integration-jdbc/blob/9e29dd625eb6c105795f15dcd1b467eee1f05453/lib/logstash/plugin_mixins/jdbc/jdbc.rb#L109-L112 👍

PR #183, opened 2025-04-01
Support other rufus scheduling options

ash-darin:
Allow jdbc input to be scheduled with different options supported by rufus scheduler.

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -180,6 +180,20 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
   # exactly once.
   config :schedule, :validate => :string
 
+  # Interval of how soon to run statement again after completion
+  # for example: "1m" (execute again 1 minute after completion)
+  #
+  # There is no interval by default. If no interval is given, then the statement is run
+  # exactly once.
+  config :interval, :validate => :string
+
+  # Start the job periodically after time elapsed
+  # for example: "1m" (execute query every minute)
+  #
+  # There is no period by default. If no period is given, then the statement is run
+  # exactly once.
+  config :period, :validate => :string
+
   # Path to file with last run time.
   # The default will write file to `<path.data>/plugins/inputs/jdbc/logstash_jdbc_last_run`
   # NOTE: it must be a file path and not a directory path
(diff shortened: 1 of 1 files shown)


robbavey:
@ash-darin Thank you for your contribution! Would you mind signing the [CLA](https://www.elastic.co/contributor-agreement)?

jsvd:
@ash-darin I'll provide that feedback to the team, but just to clarify, the initial name+email form serves only to pre-fill data in the agreement docusign, after introducing the data you get full access to the agreement and the option to start the signing process itself.

ash-darin:
@jsvd on that "next page" I have to agree "to use electronic records and signatures", marked with an "*" (that I don't know what that means or leads to), or if I want to "Finish later" I have to provide an e-mail address. Otherwise I do _not_ get access to the PDF.

jsvd:
@ash-darin right, those are docusign's own formalities on agreeing to use digital signing. if you are/your company is not comfortable agreeing to that before seeing the CLA itself I'll reach out internally to find a way to get the pdf to you first.

ash-darin:
Is there anything else I ought to do / observe?

jsvd:
Given `.every` can be implemented with a cron job (e.g. "every 3h" == `0 */3 * * *`), could we maybe add only `interval` which is different: trigger immediately and when it's done wait "interval" before doing it again?

ash-darin:
> Given `.every` can be implemented with a cron job (e.g. "every 3h" == `0 */3 * * *`), could we maybe add only `interval` which is different: trigger immediately and when it's done wait "interval" before doing it again?

`.every` ? Maybe you mistyped and meant `period`?

I don't see the reason to limit us in that way,  there is a use case for which `period` exists next to `cron` and `interval`.

`period` is slightly different from `cron`, in the sense that it runs from the moment the timer is started (logstash is started), while a cron always starts at 0 (seconds) . This difference can be advantageous if you have multiple, identically configured logstashes (as many people are prone to do with central configuration management / version management). With a cron job (and no special voodoo to randomize values) they would all trigger simultaneously (e.g. a sql query), while jobs configured with `period´ would trigger slight apart (as the logstashes would start slightly apart), though with the same repeating pattern. I had two logstashes where two cronjobs fired so precisely simultaneous that their writing to elasticsearch collided with a "version mismatch".

I would therefore ask that both, `period` and `interval` are implemented.

jsvd:
Thanks for the clarification, could I ask you to add docs for period and interval and their equivalent section to [docs/input-jdbc.asciidoc](https://github.com/logstash-plugins/logstash-integration-jdbc/blob/main/docs/input-jdbc.asciidoc)?

Example: 
* `schedule` entry in setting table https://github.com/logstash-plugins/logstash-integration-jdbc/blob/main/docs/input-jdbc.asciidoc?plain=1#L236
* `schedule` section below https://github.com/logstash-plugins/logstash-integration-jdbc/blob/main/docs/input-jdbc.asciidoc?plain=1#L565-L576

jsvd:
I'd prefer in the same PR so the same changeset introduces the feature and its associated documentation. 🙏

jsvd (approved):
CI failures are unrelated and due to an issue in the snapshot logstash docker images that will be fixed by tomorrow. LGTM. I'll merge and publish the PR. thanks for the contribution!

PR #194, opened 2026-03-26
Fix the unit test failed by jruby 10 upgrade.

mashhurs:


donoghuc (approved):
So this remains valid for both jruby 9 and 10 due to inheritance. LGTM

PR #201, opened 2026-07-17
Fix connection leak on statement retry by opening JDBC connection once outside retry loop

### LINKED ISSUE #195: Logstash JDBC input leaks Oracle sessions when a statement is cancelled and retried
**Logstash information**:

1. Logstash version: `logstash 9.2.1`
2. Logstash installation source: Docker image
3. How is Logstash being run: Kubernetes / EKS

**Plugins installed**: (`bin/logstash-plugin list --verbose`)

- logstash-integration-jdbc (5.6.1)
- logstash-input-jdbc
- logstash-filter-jdbc_streaming
- logstash-filter-jdbc_static

**JVM** (e.g. `java -version`):

- Bundled JDK in the container:
  - `openjdk version "21.0.9" 2025-10-21 LTS`
  - `OpenJDK Runtime Environment Temurin-21.0.9+10 (build 21.0.9+10-LTS)`

**OS version** (`uname -a` if on a Unix-like system):

- Linux <id>.amzn2023.x86_64 #1 SMP PREEMPT_DYNAMIC Wed May 20 06:36:23 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

**Description of the problem including expected versus actual behavior**:

When a JDBC query is cancelled by the database (for example, Oracle returns `ORA-01013: user requested cancel of current operation`), the Logstash JDBC input retries the statement. In the affected flow, the retry path opens a new database connection without first closing the previous Sequel/connection-pool state. As a result, Oracle server-side sessions remain behind as inactive sessions.

Expected behavior:
- A cancelled query should not leave orphaned database sessions behind.
- Retry logic should cleanly close the previous connection before opening a new one.

Actual behavior:
- Repeated cancellations cause Oracle sessions to accumulate as `INACTIVE` sessions.
- In a 10-cancel validation run, the number of `INACTIVE` sessions increased from 0 to 7, showing a clear leak.

This appears to be a bug in the JDBC input retry/connection lifecycle handling when the statement fails with a cancellation/timeout-like exception and the plugin retries.

**Steps to reproduce**:

1. Configure Logstash JDBC input against an Oracle 19c database with a long-running query.
2. Trigger a cancel/timeout condition on the query so Oracle returns `ORA-01013`.
3. Let the JDBC input retry the statement.
4. Observe Oracle `v$session` entries for the application user. Sessions accumulate as `INACTIVE` after repeated cancellations.

**Provide logs (if relevant)**:

Relevant log excerpts observed during the issue:

- `Java::JavaSql::SQLTimeoutException: ORA-01013: user requested cancel of current operation`
- `Unable to execute statement. Trying again.`

Observed validation result:
- 10 cancelled operations resulted in:
  - `Total=8`
  - `Active=1`
  - `Inactive=7`

This strongly indicates that the previous connection/session was not being closed cleanly during the retry flow.

alexcams:
## Problem

Fixes #195. When a JDBC query fails and is retried, `open_jdbc_connection`  is called inside the begin..retry block. Ruby's retry does not trigger ensure, so the previous connection is overwritten without being closed, leaking a database session on each retry.

## Solution

Move `open_jdbc_connection` outside the inner retry block so the connection is opened once and reused across retries. An outer begin..ensure guarantees `close_jdbc_connection` always runs, even if `open_jdbc_connection` itself fails partway through.

## How to test

Deploy a PostgreSQL (or any other DB) instance via Docker Compose (or any other method) with `statement_timeout=1000` (1 second) so any query that runs longer is cancelled by the server. Then run Logstash with a JDBC input that intentionally executes a longer query and configures multiple retry attempts:

```
input {
  jdbc {
    jdbc_driver_class      => "org.postgresql.Driver"
    jdbc_connection_string => "jdbc:postgresql://localhost:15432/testdb"
    jdbc_user              => "logstash"
    jdbc_password          => "logstash"

    # Query that sleeps 10s — will be cancelled by PostgreSQL's 1s statement_timeout
    statement => "SELECT *, pg_sleep(10) FROM test_table"

    # Force multiple retries to amplify the leak
    statement_retry_attempts           => 5
    statement_retry_attempts_wait_time => 1
  }
}

output { stdout { codec => dots } }
```

While Logstash runs, monitor sessions on the database:

```
SELECT state, count(*)
FROM pg_stat_activity
WHERE usename = 'logstash'
GROUP BY state;
```

### Expected results

- **Without the fix:** every retry opens a new connection. Idle sessions accumulate on the server (0 → 1 → 2 → 3 → 4 → 5).
- **With the fix:** the same connection is reused across all retries. Only one session exists at a time, and it is cleanly closed once the plugin finishes.

Reproduction scripts (Docker Compose file, init SQL, driver, monitor script) are available on request.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc/jdbc.rb
@@ -160,6 +160,13 @@ module LogStash  module PluginMixins module Jdbc
       logger.error('', e) # prints nested causes
     end
 
+    def log_jdbc_query_exception(e)
+      details = { exception: e.class, message: e.message }
+      details[:cause] = e.cause.inspect if e.cause
+      details[:backtrace] = e.backtrace if @logger.debug?
+      @logger.warn("Exception when executing JDBC query", details)
+    end
+
     def open_jdbc_connection
       @connection_lock.synchronize do
         # at this point driver is already loaded
@@ -233,10 +240,7 @@ module LogStash  module PluginMixins module Jdbc
             end
             success = true
           rescue Sequel::Error, Java::JavaSql::SQLException => e
-            details = { exception: e.class, message: e.message }
-            details[:cause] = e.cause.inspect if e.cause
-            details[:backtrace] = e.backtrace if @logger.debug?
-            @logger.warn("Exception when executing JDBC query", details)
+            log_jdbc_query_exception(e)
 
             if retry_attempts == 0
               @logger.error("Unable to execute statement. Tried #{@statement_retry_attempts} times.")
@@ -249,8 +253,7 @@ module LogStash  module PluginMixins module Jdbc
             @value_tracker.set_value(sql_last_value)
           end
         rescue Sequel::Error, Java::JavaSql::SQLException => e
-          # Connection-level errors from open_jdbc_connection are already logged
-          # by jdbc_connect's internal retry loop — just let ensure clean up.
+          log_jdbc_query_exception(e)
         ensure
           close_jdbc_connection
         end
--- a/lib/logstash/plugin_mixins/jdbc/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc/jdbc.rb
@@ -222,30 +222,35 @@ module LogStash  module PluginMixins module Jdbc
 
       @connection_lock.synchronize do
         begin
-          retry_attempts -= 1
           open_jdbc_connection
-          sql_last_value = @use_column_value ? @value_tracker.value : Time.now.utc
-          @tracking_column_warning_sent = false
-          @statement_handler.perform_query(@database, @value_tracker.value) do |row|
-            sql_last_value = get_column_value(row) if @use_column_value
-            yield extract_values_from(row)
-          end
-          success = true
-        rescue Sequel::Error, Java::JavaSql::SQLException => e
-          details = { exception: e.class, message: e.message }
-          details[:cause] = e.cause.inspect if e.cause
-          details[:backtrace] = e.backtrace if @logger.debug?
-          @logger.warn("Exception when executing JDBC query", details)
+          begin
+            retry_attempts -= 1
+            sql_last_value = @use_column_value ? @value_tracker.value : Time.now.utc
+            @tracking_column_warning_sent = false
+            @statement_handler.perform_query(@database, @value_tracker.value) do |row|
+              sql_last_value = get_column_value(row) if @use_column_value
+              yield extract_values_from(row)
+            end
+            success = true
+          rescue Sequel::Error, Java::JavaSql::SQLException => e
+            details = { exception: e.class, message: e.message }
+            details[:cause] = e.cause.inspect if e.cause
+            details[:backtrace] = e.backtrace if @logger.debug?
+            @logger.warn("Exception when executing JDBC query", details)
 
-          if retry_attempts == 0
-            @logger.error("Unable to execute statement. Tried #{@statement_retry_attempts} times.")
+            if retry_attempts == 0
+              @logger.error("Unable to execute statement. Tried #{@statement_retry_attempts} times.")
+            else
+              @logger.error("Unable to execute statement. Trying again.")
+              sleep(@statement_retry_attempts_wait_time)
+              retry
+            end
           else
-            @logger.error("Unable to execute statement. Trying again.")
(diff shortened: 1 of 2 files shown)


andsel:
# Test plan

## Run Dockerized MySQL

1. Create init script like, in `init-scripts/init.sql`:
  ```sql
    CREATE TABLE IF NOT EXISTS users (
        id INT AUTO_INCREMENT PRIMARY KEY,
        name VARCHAR(100) NOT NULL,
        email VARCHAR(100) NOT NULL UNIQUE,
        created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
    );
    
    INSERT INTO users (name, email) VALUES 
    ('Alice Smith', 'alice@example.com'),
    ('Bob Jones', 'bob@example.com');
  ```

2. create a `docker-compose.yml` like:
  ```yaml
    version: '3.8'
    
    services:
      db:
        image: mysql:8.0
        container_name: mysql_app_db
        restart: always
        environment:
          MYSQL_ROOT_PASSWORD: rootpassword
          MYSQL_DATABASE: my_application
          MYSQL_USER: app_user
          MYSQL_PASSWORD: app_password
        ports:
          - "3306:3306"
        volumes:
          # Mount the initialization script into the container
          - ./init-scripts:/docker-entrypoint-initdb.d
          # Optional: Named volume to persist data if the container restarts
          - mysql_data:/var/lib/mysql
    
        # THIS IS IPORTANT FOR THE TEST, must be below plugin's query sleep time
        command: --max-execution_time=2000  
    
    volumes:
      mysql_data:
  ```
3. Run the compose with:
  ```shell
    docker compose up -d
  ```

## Execute Logstash the pipeline
Download JDBC driver:

```
curl -o mysql-connector-j-8.0.33.tar.gz https://downloads.mysql.com/archives/get/p/3/file/mysql-connector-j-8.0.33.tar.gz
```
and decompres to get the jar.

Create a pipeline file like:
```
input {
  jdbc {
    jdbc_user => "app_user"
    jdbc_password => "app_password"
    jdbc_driver_class => "com.mysql.cj.jdbc.Driver"
    jdbc_driver_library => "/Users/andreaselva/workspace/logstash_tests/mysql/mysql-connector-j-8.0.33/mysql-connector-j-8.0.33.jar"
    jdbc_connection_string => "jdbc:mysql://127.0.0.1:3306/my_application"
    jdbc_validate_connection => true
    schedule => "* * * * * *"
    statement => "SELECT *, sleep(10) FROM users"
    statement_retry_attempts_wait_time => 1
    statement_retry_attempts => 5
  }
}
output {
  stdout {
    codec => rubydebug
  }  
}
```

The `sleep` time  in the  query is fundamental to be greater than DB's `max-execution_time`

Run Logstash

## Monitor the number of sessions
Use:
```shell
docker exec -it mysql_app_db mysql -u root -prootpassword -e "SHOW FULL PROCESSLIST;"
```

or 

```shell
docker exec -it mysql_app_db mysql -u root -prootpassword -e "
SELECT command, COUNT(*) FROM information_schema.processlist GROUP BY command;"
```

To check that the number of sessions still open doesn't increase.


In Logstash logs should appear the following lines:
```
[2026-07-21T16:39:16,760][ERROR][logstash.inputs.jdbc     ][main][bc18972f25dad36a5d20417d999fad9599616ef2c8db15dcf75ced80caa81a19] Unable to execute statement. Trying again.
[2026-07-21T16:39:19,771][ERROR][logstash.inputs.jdbc     ][main][bc18972f25dad36a5d20417d999fad9599616ef2c8db15dcf75ced80caa81a19] Java::JavaSql::SQLException: Query execution was interrupted, maximum statement execution time exceeded: SELECT *, sleep(10) FROM users
[2026-07-21T16:39:19,772][WARN ][logstash.inputs.jdbc     ][main][bc18972f25dad36a5d20417d999fad9599616ef2c8db15dcf75ced80caa81a19] Exception when executing JDBC query {exception: Sequel::DatabaseError, message: "Java::JavaSql::SQLException: Query execution was interrupted, maximum statement execution time exceeded", cause: "#<Java::JavaSql::SQLException: Query execution was interrupted, maximum statement execution time exceeded>"}
```

andsel (approved):
LGTM
Tested as per comment https://github.com/logstash-plugins/logstash-integration-jdbc/pull/201#issuecomment-5035497375 and works properly.

PR #203, opened 2026-08-05
Prevent concurrent Sequel JDBC subadapter initialization races by preloading adapter at driver load time

### LINKED ISSUE #202: Intermittent `NameError: uninitialized constant Sequel::JDBC::Oracle::DatabaseMethods` when multiple jdbc inputs open their first connection concurrently
**Plugin version:** logstash-integration-jdbc 5.6.3 (also applies to any version >= 5.5.2 / after [#130](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/130)), Sequel 5.96.0

**Summary**

When a single pipeline contains several `jdbc` inputs that all connect to the same database for the first time in the process, the inputs can intermittently fail at startup with `NameError: uninitialized constant Sequel::JDBC::Oracle::DatabaseMethods` (occasionally the `Sequel::JDBC::Oracle::Dataset` variant). It is timing-dependent: a subset of the inputs in the pipeline fail within a few hundred milliseconds of pipeline start, the rest succeed, and re-running the same pipeline usually works. It reproduces with the stock JRuby that ships with Logstash and is independent of the JDBC driver version (the failure is in Sequel's Ruby adapter loading, not the Java driver).

**Stack trace (trimmed)**

```
Error: uninitialized constant Sequel::JDBC::Oracle::DatabaseMethods
Did you mean?  Sequel::JDBC::Database
Exception: NameError
  org/jruby/RubyModule.java:in `const_missing'
  sequel-5.96.0/lib/sequel/adapters/jdbc/oracle.rb:11:in `block in JDBC'
  sequel-5.96.0/lib/sequel/adapters/jdbc.rb:in `adapter_initialize'
  sequel-5.96.0/lib/sequel/database/connecting.rb:in `connect'
  sequel-5.96.0/lib/sequel/core.rb:in `connect'
  logstash-integration-jdbc-5.6.3/lib/logstash/plugin_mixins/jdbc/jdbc.rb:in `jdbc_connect'
  logstash-integration-jdbc-5.6.3/lib/logstash/plugin_mixins/jdbc/jdbc.rb:in `open_jdbc_connection'
  logstash-integration-jdbc-5.6.3/lib/logstash/plugin_mixins/jdbc/jdbc.rb:in `execute_statement'
  logstash-integration-jdbc-5.6.3/lib/logstash/inputs/jdbc.rb:in `execute_query'
  logstash-integration-jdbc-5.6.3/lib/logstash/inputs/jdbc.rb:in `run'
  logstash-core/lib/logstash/java_pipeline.rb:in `inputworker'
```

The error is always raised on the run path (`execute_query` -> `open_jdbc_connection` -> `jdbc_connect` -> `Sequel.connect`), never during `register`.

**Root cause**

Sequel loads a JDBC sub-adapter lazily on the first `Sequel.connect` for a given scheme. In `sequel/adapters/jdbc/oracle.rb` the adapter proc is published into `Sequel::JDBC::DATABASE_SETUP[:oracle]` **before** the module it references (`Sequel::JDBC::Oracle::DatabaseMethods` / `Dataset`) is defined further down the same file. `Sequel::Database.load_adapter` returns that map entry to any caller the moment it exists, without waiting for the in-progress `require` to finish. So when two inputs call `Sequel.connect` for the first time at the same time, the second thread can fetch the proc and call it before the first thread has finished defining the constants, and the constant lookup raises.

Since [#130](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/130) (released in 5.5.2, see [#174](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/174)) the input opens its connection in the run phase on each input's own thread, so multiple inputs in one pipeline now do their first `Sequel.connect` concurrently. Before that change the connection was opened during the serial `register` phase, so the first-time sub-adapter load happened once, serially, and could not race. That is why this was not seen on plugin versions in the 5.4.1 - 5.5.1 range.

The plugin already recognizes this class of hazard for driver loading: `plugin_mixins/jdbc/common.rb` serializes driver jar/class loading behind a shared lock (`DRIVERS_LOADING_LOCK`, commented "to avoid concurrency related problems with multiple pipelines and multiple drivers"), but that serialization does not cover the Sequel sub-adapter load, which is what races here.

Likely the same underlying `load_adapter` code path as [#104](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/104) and [#73](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/73) (`Could not load jdbc/<db> adapter: adapter class not registered in ADAPTER_MAP`) and related to the concurrent-driver-loading discussion in [#99](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/99).

**Reproduction**

Point several inputs at the same Oracle database in one pipeline so they all do their first connect at once, and run repeatedly (it is intermittent, so loop it). Any Oracle target and driver work; the driver version does not matter.

```
input {
  jdbc {
    jdbc_driver_library => "/path/to/ojdbc8.jar"
    jdbc_driver_class => "Java::oracle.jdbc.driver.OracleDriver"
    jdbc_connection_string => "jdbc:oracle:thin:@//host:1521/SERVICE"
    jdbc_user => "user"
    jdbc_password => "pass"
    statement => "select 1 from dual"
  }
  jdbc {
    jdbc_driver_library => "/path/to/ojdbc8.jar"
    jdbc_driver_class => "Java::oracle.jdbc.driver.OracleDriver"
    jdbc_connection_string => "jdbc:oracle:thin:@//host:1521/SERVICE"
    jdbc_user => "user"
    jdbc_password => "pass"
    statement => "select 2 from dual"
  }
  # ... repeat for ~10 inputs to widen the race window
}
output { stdout { codec => dots } }
```

```
while true; do bin/logstash -f repro.conf --path.data /tmp/ls-$RANDOM -w 5; done
```

A run that hits the race logs the `NameError` for a subset of the inputs shortly after `Pipeline started`; the same config re-run usually succeeds.

**Suggested direction**

Load the Sequel JDBC sub-adapter once, up front, before the concurrent run-phase connects, so its first-time `require` cannot race. The natural place is alongside the existing serial driver loading in `common.rb#load_driver` (which runs during `register` and already holds `DRIVERS_LOADING_LOCK`), loading the sub-adapter for the connection string's scheme the same way Sequel does internally (`Sequel::Database.load_adapter(scheme, map: Sequel::JDBC::DATABASE_SETUP, subdir: 'jdbc')`). This loads adapter code only and does **not** open a database connection, so it does not reintroduce the register-phase connection behavior that [#130](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/130) / [#174](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/174) deliberately moved out of `register`. It is also driver-agnostic, so it would cover the mssql/sybase variants in [#104](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/104) / [#73](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/73) / [#99](https://github.com/logstash-plugins/logstash-integration-jdbc/issues/99).

There is also a deeper fix available upstream in Sequel (define the sub-adapter module before publishing its `DATABASE_SETUP` entry, or have `load_adapter` not return an entry whose `require` is still in flight), but the plugin-side preload stands on its own and needs no Sequel release.

alexcams:
## Summary
Fixes an intermittent startup race where multiple `jdbc` inputs connecting concurrently can fail with:

- `NameError: uninitialized constant Sequel::JDBC::Oracle::DatabaseMethods`

Fixes #202.

## Problem
Since connections are opened in the input run phase (per-input thread), several `jdbc` inputs in the same pipeline can call `Sequel.connect` for the same JDBC scheme at nearly the same time on first use.

Sequel lazily loads JDBC subadapters (`jdbc/oracle`, `jdbc/mssql`, etc.). During that first-load window, another thread can observe partially initialized subadapter state and hit a `NameError` while constants are still being defined.

This is timing-dependent and intermittent, which is why reruns often succeed.

## Solution applied
During `load_driver` (register-time path already protected by `DRIVERS_LOADING_LOCK`), preload the Sequel JDBC subadapter derived from the JDBC connection string scheme.

Implementation details:
- Parse subadapter scheme from the JDBC URL (e.g. `jdbc:oracle:...` -> `:oracle`).
- Call Sequel adapter preload with:
  - `Sequel::Database.load_adapter(subadapter, map: Sequel::JDBC::DATABASE_SETUP, subdir: 'jdbc')`
- Keep behavior tolerant when no dedicated subadapter is available:
  - rescue `Sequel::AdapterNotFound` and continue (generic JDBC path still works for some DBs).

This preserves the intended behavior of not opening DB connections during register while ensuring first-time adapter code loading cannot race in run-phase worker threads.

## Manual testing
Manual repro followed the issue’s reproduction approach (multiple jdbc inputs to same Oracle DB, repeated fresh starts), with an amplified stress scenario.

### Setup
- Oracle DB via Docker (`gvenzl/oracle-free:23-slim`)
- Oracle JDBC driver jar downloaded and used by pipeline
- Generated stress pipeline with 40 concurrent `jdbc` input blocks (same DB, different `select <n> from dual` statements)
- Logstash started repeatedly with fresh `--path.data`

### Reproduce on pre-fix state
Using the same stress scenario:
- 40 inputs, 8 workers
- repeated runs until hit
- reproduced on run 12
- observed multiple occurrences of:
  - `uninitialized constant Sequel::JDBC::Oracle::DatabaseMethods`
  - `Exception: NameError`

### Verify on fixed state
Same scenario, same environment:
- 40 inputs, 8 workers
- 100-run stress pass: 0 `NameError` hits

Result: the race reproduces before the change and is not observed after the change under identical stress conditions.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/common.rb
+++ b/lib/logstash/plugin_mixins/jdbc/common.rb
@@ -32,6 +32,7 @@ module LogStash module PluginMixins module Jdbc
       # concurrency related problems with multiple pipelines and multiple drivers
       DRIVERS_LOADING_LOCK.lock()
       begin
+        preload_sequel_jdbc_subadapter
         load_driver_jars
         begin
           @driver_impl = load_jdbc_driver_class
@@ -48,6 +49,23 @@ module LogStash module PluginMixins module Jdbc
       ensure
         DRIVERS_LOADING_LOCK.unlock()
       end
+    end
+
+    def preload_sequel_jdbc_subadapter
+      subadapter = jdbc_subadapter_scheme
+      return unless subadapter
+
+      # Ensure first-time Sequel jdbc/<subadapter> loading happens serially.
+      Sequel::Database.load_adapter(subadapter, :map => Sequel::JDBC::DATABASE_SETUP, :subdir => 'jdbc')
+    rescue Sequel::AdapterNotFound
+      # Some JDBC URLs can work without a dedicated Sequel sub-adapter.
+      nil
+    end
+
+    def jdbc_subadapter_scheme
+      return nil unless @jdbc_connection_string
+
+      @jdbc_connection_string[/\Ajdbc:([^:]+):/, 1]&.to_sym
     end
 
     def load_driver_jars
--- a/lib/logstash/plugin_mixins/jdbc/common.rb
+++ b/lib/logstash/plugin_mixins/jdbc/common.rb
@@ -65,7 +65,7 @@ module LogStash module PluginMixins module Jdbc
     def jdbc_subadapter_scheme
       return nil unless @jdbc_connection_string
 
-      @jdbc_connection_string[/\Ajdbc:([^:]+):/, 1]&.to_sym
+      @jdbc_connection_string[/\Ajdbc:([^:]+):/, 1]&.downcase&.to_sym
     end
 
     def load_driver_jars
--- a/lib/logstash/plugin_mixins/jdbc/common.rb
+++ b/lib/logstash/plugin_mixins/jdbc/common.rb
@@ -59,6 +59,7 @@ module LogStash module PluginMixins module Jdbc
       Sequel::Database.load_adapter(subadapter, :map => Sequel::JDBC::DATABASE_SETUP, :subdir => 'jdbc')
     rescue Sequel::AdapterNotFound
       # Some JDBC URLs can work without a dedicated Sequel sub-adapter.
+      @logger.debug("Skipping Sequel JDBC sub-adapter preload", :subadapter => subadapter, :jdbc_connection_string => @jdbc_connection_string)
       nil
     end
 
--- a/spec/inputs/jdbc_spec.rb
+++ b/spec/inputs/jdbc_spec.rb
@@ -1913,10 +1913,24 @@ describe LogStash::Inputs::Jdbc do
       allow(plugin).to receive(:load_jdbc_driver_class).and_return(double("driver_class"))
     end
 
+    it "extracts jdbc sub-adapter scheme in lowercase" do
+      plugin.instance_variable_set(:@jdbc_connection_string, "jdbc:Oracle:thin:@//localhost:1521/FREEPDB1")
+
+      expect(plugin.send(:jdbc_subadapter_scheme)).to eq(:oracle)
+    end
+
     it "preloads sequel jdbc sub-adapter for jdbc URLs" do
       expect(Sequel::Database).to receive(:load_adapter).with(:derby, :map => Sequel::JDBC::DATABASE_SETUP, :subdir => 'jdbc').and_call_original
 
       plugin.send(:load_driver)
+    end
+
+    it "uses normalized jdbc sub-adapter scheme while preloading" do
+      plugin.instance_variable_set(:@jdbc_connection_string, "jdbc:Oracle:thin:@//localhost:1521/FREEPDB1")
+
+      expect(Sequel::Database).to receive(:load_adapter).with(:oracle, :map => Sequel::JDBC::DATABASE_SETUP, :subdir => 'jdbc').and_return(nil)
+
+      expect { plugin.send(:load_driver) }.not_to raise_error
     end
 
     it "does not try to preload when URL is not jdbc" do
--- a/spec/inputs/jdbc_spec.rb
+++ b/spec/inputs/jdbc_spec.rb
@@ -1907,4 +1907,40 @@ describe LogStash::Inputs::Jdbc do
       end
     end
   end
+
+  describe "jdbc adapter preloading", :no_connection do
+    before do
+      allow(plugin).to receive(:load_jdbc_driver_class).and_return(double("driver_class"))
+    end
+
+    it "preloads sequel jdbc sub-adapter for jdbc URLs" do
+      expect(Sequel::Database).to receive(:load_adapter).with(:derby, :map => Sequel::JDBC::DATABASE_SETUP, :subdir => 'jdbc').and_call_original
+
+      plugin.send(:load_driver)
+    end
+
+    it "does not try to preload when URL is not jdbc" do
+      plugin.instance_variable_set(:@jdbc_connection_string, "mock://localhost:1527/db")
+
+      expect(Sequel::Database).not_to receive(:load_adapter)
+
+      plugin.send(:load_driver)
+    end
+
+    it "ignores unknown jdbc sub-adapters" do
+      plugin.instance_variable_set(:@jdbc_connection_string, "jdbc:unknown://localhost:1527/db")
+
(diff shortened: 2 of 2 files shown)


alexcams:
/pr_prepare release:patch
- Prevent concurrent Sequel JDBC subadapter initialization races by preloading adapter at driver load time

lib/logstash/plugin_mixins/jdbc/common.rb
```diff
@@ -50,6 +51,23 @@ def load_driver
...
+      Sequel::Database.load_adapter(subadapter, :map => Sequel::JDBC::DATABASE_SETUP, :subdir => 'jdbc')
+    rescue Sequel::AdapterNotFound
+      # Some JDBC URLs can work without a dedicated Sequel sub-adapter.
+      nil
+    end
+
+    def jdbc_subadapter_scheme
+      return nil unless @jdbc_connection_string
+
+      @jdbc_connection_string[/\Ajdbc:([^:]+):/, 1]&.to_sym
```
donoghuc: Is there a reason this is case sensitive?

lib/logstash/plugin_mixins/jdbc/common.rb
```diff
@@ -50,6 +51,23 @@ def load_driver
...
     end
 
+    def preload_sequel_jdbc_subadapter
+      subadapter = jdbc_subadapter_scheme
+      return unless subadapter
+
+      # Ensure first-time Sequel jdbc/<subadapter> loading happens serially.
+      Sequel::Database.load_adapter(subadapter, :map => Sequel::JDBC::DATABASE_SETUP, :subdir => 'jdbc')
+    rescue Sequel::AdapterNotFound
+      # Some JDBC URLs can work without a dedicated Sequel sub-adapter.
```
donoghuc, line 61: Can we log a debug message that this happened?

donoghuc (changes requested):
Overall i think this approach will work. Couple small clarifying questions before approval.

lib/logstash/plugin_mixins/jdbc/common.rb
alexcams: It wasn't deliberately, the regex just extracts the subadapter token as it is. I fixed this in [7419927](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/203/commits/741992764be46eb6a2bac5afe4ab6f0e3eaeab46). Thanks for notice!

lib/logstash/plugin_mixins/jdbc/common.rb
alexcams: Yes, added in [87c391a](https://github.com/logstash-plugins/logstash-integration-jdbc/pull/203/commits/87c391a5b60084d2de48039d8796e6140747e1c4)

donoghuc (approved):
sweet. thanks for addressing the questions.

alexcams:
Thanks @donoghuc for the review and the detailed issue description. It really helped to reproduce the bug.

